PR #21, opened 2016-10-31
Do not merge - Fix performance issues with member registration

SFWMuhammadI:
- Add missing indexes for foreign keys
- Query as no tracking to improve performance


SFWMuhammadI:
Changes will be made in a feature branch instead

PR #23, opened 2016-11-01
#54610 - View summary totals per submission

SFWMuhammadI:


src/EA.Weee.Web/Areas/Admin/Controllers/SubmissionsController.cs
```diff
@@ -247,8 +248,10 @@ public SubmissionsController(BreadcrumbService breadcrumb, Func<IWeeeClient> cli
                 try
                 {
                     var schemeData = await client.SendAsync(User.GetAccessToken(), new GetSchemeById(schemeId));
+                    var getDataReturnSubmissionsHistoryResults = new GetDataReturnSubmissionsHistoryResults(schemeId, schemeData.OrganisationId, year);
+                    getDataReturnSubmissionsHistoryResults.IncludeSummaryData = true;
```
SFWChrisSmith: Should only ever be setting this in the constructor. It does not change after the request creation.

PR #25, opened 2016-11-02
Feature/update nu get packages

SFWChrisSmith:


src/EA.Weee.Api/Logging/DebugLogSink.cs
```diff
@@ -0,0 +1,27 @@
...
+        private readonly MessageTemplateTextFormatter formatter;
+
+        public DebugLogSink(MessageTemplateTextFormatter formatter)
+        {
+            this.formatter = formatter;
+        }
+
+        public void Emit(LogEvent logEvent)
+        {
+            var sr = new StringWriter();
```
SFWMuhammadI: May be worth wrapping this with a using statement

PR #74, opened 2019-02-04
Feature/summaryscreen

SamGoodeySFW:


src/EA.Weee.Web/Areas/AATFReturn/Controllers/CheckYourReturnController.cs
```diff
@@ -0,0 +1,87 @@
+﻿namespace EA.Weee.Web.Areas.AatfReturn.Controllers
```
JohnPadgettCivica: The wiew model should build the totals and I think the API call should do more than just return a list of non obligated values it should return an entire return data set (i.e. what will be there going forward (think all of the AATF data etc), this data object can be passed to a mapper function that will create the view model i.e. where the totals display should be calculated...  Have a look at the submiited return data access and view models to see what can be re-used or expanded on.

PR #80, opened 2019-02-08
Feature/selectpcs

SamGoodeySFW:
Created operator data class to enable ReturnData class to hold information about it's relevant operator


src/EA.Weee.RequestHandlers.Tests.Unit/Mapping/ReturnMapTests.cs
```diff
@@ -33,7 +34,7 @@ public void Map_GivenSourceIsNull_ArgumentNullExceptionExpected()
         [Fact]
         public void Map_GivenSource_QuarterPropertiesShouldBeMapped()
         {
```
JohnPadgettCivica: Can you add a test to make sure that the operator properties are mapped

PR #95, opened 2019-02-19
Feature/showweeereceivedvalues

JB12298:
Story 67316


src/EA.Weee.Web/Areas/AATFReturn/ViewModels/NonObligatedValuesViewModel.cs
```diff
@@ -17,14 +18,18 @@ public class NonObligatedValuesViewModel
 
         public bool Dcf { get; set; }
 
+        private ICategoryValueTotalCalculator categoryValueCalculator;
+
         public NonObligatedValuesViewModel()
```
JohnPadgettCivica: Should be able to pass ICategoryValueTotalCalculator as constructor parameter rather than instantiate in the constructor

PR #132, opened 2019-03-15
Feature/weee sent on site bug fixes

SamGoodeySFW:


src/EA.Weee.Core.Tests.Unit/AatfReturn/AddressDataTests.cs
```diff
@@ -13,9 +13,9 @@ public class AddressDataTests
         [InlineData("Address1")]
```
JohnPadgettCivica: Can we rename this file and tests to Site address tests and also create the same tests for the Aatf address

PR #135, opened 2019-03-19
Feature/reuse site update

JB12298:


JohnPadgettCivica:
Just need to merge from develop and then can finish this one

PR #142, opened 2019-03-25
Feature/weee sent on operator

SamGoodeySFW:


src/EA.Weee.Web/Areas/AATFReturn/Controllers/SentOnCreateSiteOperatorController.cs
```diff
@@ -0,0 +1,107 @@
...
+
+                var request = getRequestCreator.ViewModelToRequest(viewModel);
+
+                var siteAddressData = await client.SendAsync(User.GetAccessToken(), request);
+
+                viewModel.SiteAddressData = siteAddressData;
+
+                if (javascriptDisabled != null)
+                {
+                    var disabled = (bool)javascriptDisabled;
```
JohnPadgettCivica: Do this mapping in the mapper class. Pass the javascriptDisabled action parameter in with the transfer object and do the check and setting of the operator address fields.

Or probably better... have a site address property on ReturnAndAatfToSentOnCreateSiteOperatorViewModelMapTransfer. If javascript is disabled retrieve the address and pass the address on the new property. In the mapper check that property, if its not null then set the address fields to the one passed in values.

src/EA.Weee.Web/Areas/AATFReturn/Views/SentOnCreateSiteOperator/Index.cshtml
```diff
@@ -0,0 +1,112 @@
...
+
+                if (address1 != '') {
+                    $('#OperatorAddressData_Address1').val(address1);
+                }
+
+                if (address2 != '') {
+                    $('#OperatorAddressData_Address2').val(address2);
+                }
+
+                if (townorcity != '') {
```
JohnPadgettCivica: Put this javascript into the sent-on-operator-address.js file and include on this page so it gets minified

src/EA.Weee.Web/Areas/AATFReturn/Controllers/SentOnCreateSiteOperatorController.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                AatfAddressData siteAddressDataJSDisabled = new AatfAddressData();
+
+                if (javascriptDisabled == true)
+                {
+                    var requestJSDisabled = new GetSentOnAatfSite(weeeSentOnId);
+
+                    siteAddressDataJSDisabled = await client.SendAsync(User.GetAccessToken(), requestJSDisabled);
+                }
+
```
JohnPadgettCivica: Do this before mapper.Map and pass the siteAddress into the mapper, so we don't need to  do viewModel.SiteAddressData = siteAddressData; in the action... Also unit tests for this and the mapper.

You may need to create a new property on the view model for the address transfer.

PR #169, opened 2019-04-16
added in copy and paste for obligated sent on

willdot:
PBI: 69684

Added the copy paste functionality to the obligated sent on for AATFs

### DIFF
--- a/src/EA.Weee.Web/Areas/AatfReturn/Controllers/ObligatedSentOnController.cs
+++ b/src/EA.Weee.Web/Areas/AatfReturn/Controllers/ObligatedSentOnController.cs
@@ -2,8 +2,8 @@
 {
     using EA.Prsd.Core.Mapper;
     using EA.Weee.Api.Client;
+    using EA.Weee.Core.AatfReturn;
     using EA.Weee.Requests.AatfReturn;
-    using EA.Weee.Requests.AatfReturn.Obligated;
     using EA.Weee.Web.Areas.AatfReturn.Mappings.ToViewModel;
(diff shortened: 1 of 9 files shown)


JohnPadgettCivica:
is it possible to re-use the ObligatedValuesCopyPasteController and add a redirect so we know where to redirect back to? Will make sure that if we need to make changes to this copy and paste we only need to make it in the one place...

willdot:
> is it possible to re-use the ObligatedValuesCopyPasteController and add a redirect so we know where to redirect back to? Will make sure that if we need to make changes to this copy and paste we only need to make it in the one place...

Attempted that yesterday, but got messy because the different controllers accept different parameters to the action methods.

src/EA.Weee.Web/Areas/AatfReturn/Controllers/ObligatedSentOnController.cs
```diff
@@ -50,7 +50,8 @@ public virtual async Task<ActionResult> Index(Guid returnId, Guid organisationId
                     ReturnData = @return,
                     AatfId = aatfId,
                     OperatorName = operatorName,
-                    WeeeSentOnId = weeeSentOnId
+                    WeeeSentOnId = weeeSentOnId,
+                    PastedData = TempData["pastedValues"] as ObligatedCategoryValue
```
JohnPadgettCivica: Can we add a unit test to make sure that if the pasted values are provided they are set on the mapper object? ta...

willdot: Just added a test.

PR #176, opened 2019-04-18
Feature/chk return new layout

jaswinder2303:


src/EA.Weee.Web/Areas/AatfReturn/Views/CheckYourReturn/Index.cshtml
```diff
@@ -256,26 +261,71 @@
...
             </div>
         </div>
-    </div>
-</section>
\ No newline at end of file
+        </div>
+</section>
+
+<script type="text/javascript">
+    document.getElementById('collapsibleHref').onclick = function() {
```
SamGoodeySFW: Extract this JS out into it's own scripts, add it to a bundle in Web\App_Start\BundleConfig, then render it on this page. Similar to to how it's done on SentOnCreateSiteOperator Index page and the sent-on-operator-address JS script

PR #184, opened 2019-04-26
Feature/report select

JB12298:


JB12298:
@SamGoodeySFW @jaswinder2303 

- Added Mapper & Tests
- Fixed magic string "Yes" in request handler
- Added View Model test
- Added Integration test
- Added a GetAll data access test
- Added authorisation to GetReportOnQuestionHandler & Tests

I think that's what was covered in the review?  Let me know if I've missed something

PR #207, opened 2019-05-03
changed to show org name when confirming

willdot:
Task: 70322


JohnPadgettCivica:
This one seems to have been missed, if you merge from develop ill approive this one...

PR #208, opened 2019-05-03
redirect user to the aatf list if there are tonnage already entered

willdot:
Task: 70249

### DIFF
--- a/src/EA.Weee.Web/Areas/AatfReturn/Views/AatfTaskList/Index.cshtml
+++ b/src/EA.Weee.Web/Areas/AatfReturn/Views/AatfTaskList/Index.cshtml
@@ -80,13 +80,13 @@
         <div class="govuk-grid-row govuk-table__cell weee-tonnage-top-aligned-display">
             <div class="govuk-grid-column-one-half">
                 <p>
-                    @if (Equals(item.WeeeSentOn.B2B, "-") && Equals(item.WeeeSentOn.B2C, "-"))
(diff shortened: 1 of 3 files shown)


src/EA.Weee.Web/Areas/AatfReturn/Views/AatfTaskList/Index.cshtml
```diff
@@ -80,7 +80,15 @@
         <div class="govuk-grid-row govuk-table__cell weee-tonnage-top-aligned-display">
             <div class="govuk-grid-column-one-half">
                 <p>
-                    @Html.NavigationRouteLink("Sent to another AATF / ATF", @item.Aatf.Name + " sent to another AATF / ATF", AatfRedirect.AatfSelectedRoute, new { returnId = Model.ReturnId, aatfId = @item.Aatf.Id, action = "Index", controller = "SentOnCreateSite" })
+                    @if (Equals(item.WeeeSentOn.B2B, "-") && Equals(item.WeeeSentOn.B2C, "-"))
```
JohnPadgettCivica: So we can better test this, is it possible to add a property on the AatfsData item say bool RedirectSummary which set in the view model mapper, we can then create a unit test? Then use that bool value as the if statement.

Doing this in the View makes it harder to test.

willdot: Done the refactoring and added a test.

Had to slightly change the test data to use the same Guid as an ID because things weren't matching up.

PR #211, opened 2019-05-03
Feature/tasklist options

JB12298:


JohnPadgettCivica:
Update the HandleAsync_GivenReturn_MapperShouldBeCalled to test for the reports on

src/EA.Weee.Web/Areas/AATFReturn/Controllers/SelectReportOptionsController.cs
```diff
@@ -71,7 +72,7 @@ public virtual async Task<ActionResult> Index(SelectReportOptionsViewModel viewM
 
             if (ModelState.IsValid)
             {
-                if (viewModel.SelectedOptions != null)
+                if (viewModel.SelectedOptions != null && viewModel.SelectedOptions.Count != 0)
```
JohnPadgettCivica: Can add a property on the view model HasSelectedOptions that would do the null and count checks

PR #216, opened 2019-05-08
Bug/misc

jaswinder2303:


jaswinder2303:
added a check

src/EA.Weee.Web/Areas/AatfReturn/Views/NonObligatedValuesCopyPaste/Index.cshtml
```diff
@@ -8,7 +8,10 @@
 <section id="main-content">
     <header class="hgroup">
         <div>
-            <h1 id="data-entry-label" class="govuk-heading-l">Non-obligated WEEE</h1>
+            @{
+                var typeheading = Model.Dcf == false ? "Non-obligated WEEE" :"Non-obligated WEEE retained by a DCF";
```
SamGoodeySFW: Sorry to be a pain, would it be possible to move this calculation into the controller and assign typeheading to the view model? Just so we don't have code split between the controller and the index

jaswinder2303: moved the code.

PR #218, opened 2019-05-08
Feature/submitted returns

JohnPadgettCivica:
Product Backlog Item 69719:View previously submitted returns / Manage AATF returns screen


src/EA.Weee.Domain/AatfReturn/Return.cs
```diff
@@ -25,7 +25,7 @@ public Return(Operator aatfOperator, Quarter quarter, string createdBy)
             CreatedDate = SystemTime.UtcNow;
         }
 
-        public void UpdateSubmitted(string submittedBy)
+        public virtual void UpdateSubmitted(string submittedBy)
```
willdot: Just a curious question. If this is not being over ridden anywhere, why is it virtual?

PR #221, opened 2019-05-10
Feature/edit aatf contact

JB12298:


src/EA.Weee.Database/scripts/Update/Release 2/Sprint8/20190508-0900-UpdateAATFTable.sql
```diff
@@ -0,0 +1,42 @@
...
+           ,[Postcode]
+           ,[CountryId]
+           ,[Telephone]
+           ,[Email])
+     VALUES
+           (@defaultContactId
+           ,'First Name'
+           ,'Last Name'
+           ,'Director'
+           ,'1 Address Lane'
```
JohnPadgettCivica: Remove the transaction so we don't have same issue as Will's :)

src/EA.Weee.Domain.Tests.Unit/AatfReturn/AatfContactTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            Action constructor = () =>
+            {
+                var @return = new AatfContact(firstName, lastName, position, address, A.Dummy<string>(), town, A.Dummy<string>(), A.Dummy<string>(), A.Dummy<Country>(), telephone, email);
+            };
+
+            constructor.Should().Throw<ArgumentException>();
+        }
+    }
+}
```
JohnPadgettCivica: Add test to check that the properties passed are mapped.

Possible to aldo add test for the update contact method.

src/EA.Weee.RequestHandlers/AatfReturn/Internal/GetAatfContactHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            authorization.EnsureCanAccessInternalArea();
+
+            var contact = await dataAccess.GetContact(message.AatfId);
+
+            var result = mapper.Map(contact);
+
+            result.CanEditContactDetails = authorization.CheckUserInRole(Roles.InternalAdmin);
+
+            return result;
```
JohnPadgettCivica: Add Test to check that the result is the result of the mapped object
Add Test to check that the CanEditContactDetails property is set based on the authorization.CheckUserInRole

src/EA.Weee.Web/Areas/Admin/Controllers/AatfController.cs
```diff
@@ -76,6 +83,56 @@ public async Task<ActionResult> ManageAatfs(ManageAatfsViewModel viewModel)
...
+
+                    await client.SendAsync(User.GetAccessToken(), request);
+
+                    return RedirectToAction("Details", new { Id = viewModel.AatfId });
+                }
+            }
+
+            using (var client = apiClient())
+            {
+                viewModel.ContactData.AddressData.Countries = await client.SendAsync(User.GetAccessToken(), new GetCountries(false));
```
JohnPadgettCivica: Not sure we have done this elsewhere, but if we have to re get the countries list can we create a unit test to make sure the model property is set based from the api return

src/EA.Weee.Web/Areas/Admin/Mappings/ToViewModel/AatfDataToAatfDetailsViewModel.cs
```diff
@@ -3,29 +3,40 @@
     using EA.Prsd.Core;
     using EA.Prsd.Core.Mapper;
     using EA.Weee.Core.AatfReturn;
+    using EA.Weee.Web.Areas.AatfReturn.Mappings.ToViewModel;
     using EA.Weee.Web.Areas.Admin.ViewModels.Aatf;
     using System;
 
-    public class AatfDataToAatfDetailsViewModel : IMap<AatfData, AatfDetailsViewModel>
+    public class AatfDataToAatfDetailsViewModel : IMap<AatfDataToAatfDetailsViewModelTransfer, AatfDetailsViewModel>
     {
```
JohnPadgettCivica: Rename to AatfDataToAatfDetailsViewModelMapper

src/EA.Weee.Web/Areas/Admin/Controllers/AatfController.cs
JB12298: I don't check this specifically, but I check that countries request is made?  AatfControllerTests line 211

JohnPadgettCivica: Cool, if you can check that the API call is made and the view model contains the value returns from the api...

JB12298: Added a test

PR #229, opened 2019-05-13
fixed error where on reload or back after deleting, an error messagle…

willdot:
…d. Now the user is redirected. Also fixed the wrong parameter order.

Task: 70825

The order of parameters on the redirect kept telling me the user doesn't have permission to the organisation...but that's because the ID was not for an organisation :D

### DIFF
--- a/src/EA.Weee.RequestHandlers/AatfReturn/ObligatedSentOn/GetWeeeSentOnByIdHandler.cs
+++ b/src/EA.Weee.RequestHandlers/AatfReturn/ObligatedSentOn/GetWeeeSentOnByIdHandler.cs
@@ -36,6 +36,11 @@
             var weeeSentOnList = new List<WeeeSentOnData>();
             var weeeSentOn = await getSentOnAatfSiteDataAccess.GetWeeeSentOnById(message.WeeeSentOnId);
 
+            if (weeeSentOn == null)
+            {
+                return null;
+            }
+
             var amount = await fetchWeeeSentOnAmountDataAccess.FetchObligatedWeeeSentOnForReturn(weeeSentOn.Id);
 
             var weeeSentOnObligatedData = amount.Select(n => new WeeeObligatedData(n.Id, new AatfData(n.WeeeSentOn.Aatf.Id, n.WeeeSentOn.Aatf.Name, n.WeeeSentOn.Aatf.ApprovalNumber), n.CategoryId, n.NonHouseholdTonnage, n.HouseholdTonnage)).ToList();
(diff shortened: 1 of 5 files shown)


src/EA.Weee.RequestHandlers/AatfReturn/ObligatedSentOn/GetWeeeSentOnByIdHandler.cs
```diff
@@ -36,6 +36,11 @@ public async Task<WeeeSentOnData> HandleAsync(GetWeeeSentOnById message)
             var weeeSentOnList = new List<WeeeSentOnData>();
             var weeeSentOn = await getSentOnAatfSiteDataAccess.GetWeeeSentOnById(message.WeeeSentOnId);
 
+            if (weeeSentOn == null)
+            {
+                return null;
```
SamGoodeySFW: Unit test to make sure handle async returns null when weeeSentOn is null

src/EA.Weee.Web/Areas/AATFReturn/Controllers/SentOnRemoveSiteController.cs
```diff
@@ -40,7 +40,13 @@ public virtual async Task<ActionResult> Index(Guid organisationId, Guid returnId
             {
                 var weeeSentOn = await client.SendAsync(User.GetAccessToken(), new GetWeeeSentOnById(weeeSentOnId));
 
+                if (weeeSentOn == null)
+                {
+                    return AatfRedirect.SentOnSummaryList(organisationId, returnId, aatfId);
```
SamGoodeySFW: Unit test to make sure redirect is correct when weeeSentOn is null

src/EA.Weee.Web/Areas/AATFReturn/Controllers/SentOnRemoveSiteController.cs
```diff
@@ -71,7 +77,7 @@ public virtual async Task<ActionResult> Index(SentOnRemoveSiteViewModel viewMode
                     {
                         var weeeSentOn = await client.SendAsync(User.GetAccessToken(), new RemoveWeeeSentOn(viewModel.WeeeSentOnId));
                     }
-                    return AatfRedirect.SentOnSummaryList(viewModel.ReturnId, viewModel.AatfId, viewModel.OrganisationId);
+                    return AatfRedirect.SentOnSummaryList(viewModel.OrganisationId, viewModel.ReturnId, viewModel.AatfId);
                 }
             }
```
JB12298: Is there a test to check the return View(viewModel);

PR #239, opened 2019-05-15
fixed alignment on IE and Edge

willdot:
As seen in demo. Damn IE and Edge.....

Also fixed Appropriate typo


JohnPadgettCivica:
Assume still works in chrome? :)

willdot:
> Assume still works in chrome? :)

It does. Haven't tested firefox though as I don't have it and always fails to install on my laptop.

PR #241, opened 2019-05-17
Feature/create aatf

willdot:
PBI: 70849


src/EA.Weee.RequestHandlers/Mappings/AatfAddressDataMap.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace EA.Weee.RequestHandlers.Mappings
+{
+    using EA.Weee.Core.AatfReturn;
+    using EA.Weee.Domain;
+    using EA.Weee.Domain.AatfReturn;
+    using Prsd.Core.Mapper;
+
+    public class AatfAddressDataMap : IMap<AatfAddressData, AatfAddress>
+    {
+        public AatfAddress Map(AatfAddressData source)
```
JB12298: Please add unit tests

src/EA.Weee.RequestHandlers/Mappings/AatfContactDataMap.cs
```diff
@@ -0,0 +1,33 @@
...
+    using EA.Prsd.Core.Mapper;
+    using EA.Weee.Core.AatfReturn;
+    using EA.Weee.DataAccess;
+    using EA.Weee.Domain;
+    using EA.Weee.Domain.AatfReturn;
+    using System.Linq;
+
+    public class AatfContactDataMap : IMap<AatfContactData, AatfContact>
+    {
+        private readonly WeeeContext context;
```
JB12298: Unit tests :)

src/EA.Weee.RequestHandlers/Mappings/UKCompetentAuthorityDataMap.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    using Core.Shared;
+    using Domain;
+    using EA.Weee.DataAccess;
+    using Prsd.Core.Mapper;
+    using System.Linq;
+
+    public class UKCompetentAuthorityDataMap : IMap<UKCompetentAuthorityData, UKCompetentAuthority>
+    {
+        private readonly WeeeContext context;
```
JB12298: Unit Tests

src/EA.Weee.Web/Areas/Admin/Controllers/AddAatfController.cs
```diff
@@ -55,7 +69,77 @@ public async Task<ActionResult> SearchResults(SearchResultsViewModel viewModel)
                 return View(viewModel);
```
JB12298: I think it's missing a unit test for returning the view

willdot: Was calling get method instead of post in one of the tests. Now fixed.

src/EA.Weee.RequestHandlers/Mappings/AatfAddressDataMap.cs
willdot: Done

src/EA.Weee.RequestHandlers/Mappings/AatfContactDataMap.cs
willdot: Done

src/EA.Weee.RequestHandlers/Mappings/UKCompetentAuthorityDataMap.cs
willdot: Hard to do as it needs a WeeeContext, which can't be mocked or faked because it's not an interface. Will discuss with John on Monday

src/EA.Weee.DataAccess.Tests.Integration/ObligatedWeeeIntegrationCommon.cs
```diff
@@ -31,7 +31,8 @@ public static Return CreateReturn(Operator operatorTest, string createdById)
 
         public static Aatf CreateAatf(UKCompetentAuthority competentAuthority, Operator @operator, AatfContact contact)
         {
-            var aatf = new Aatf("aatfname", competentAuthority, "number", AatfStatus.Approved, @operator, CreateAatfAddress(null), AatfSize.Large, DateTime.Now, contact);
+            Domain.Country country = new Domain.Country(Guid.NewGuid(), "Uk");
```
JohnPadgettCivica: Don't do new Country - permissions on SQL server will mean this will fail in other environments - select the country as has been done in other places.

src/EA.Weee.RequestHandlers/Mappings/AatfContactDataMap.cs
```diff
@@ -0,0 +1,26 @@
...
+    using EA.Weee.Domain.AatfReturn;
+    using System.Linq;
+
+    public class AatfContactDataMap : IMap<AatfContactData, AatfContact>
+    {
+        public AatfContact Map(AatfContactData source)
+        {
+            if (source != null)
+            {
+                Country country = new Country(source.AddressData.CountryId, source.AddressData.CountryName);
```
JohnPadgettCivica: Here pass the country in via the AatfContactData source, we can't do a new.

willdot: I've realized that I don't need a country object to create an AATF, so just using the country Id that is already passed in.

src/EA.Weee.DataAccess.Tests.Integration/ObligatedWeeeIntegrationCommon.cs
willdot: Ok fixed where found.

src/EA.Weee.RequestHandlers/Mappings/UKCompetentAuthorityDataMap.cs
willdot: Realised that don't need a whole Competent auth object, just the ID so this mapper is no longer needed

JohnPadgettCivica: Just for information - We can fake the WeeeContext and set the data sets to return test lists, an example is in GetOrAddAatfDeliveryLocation_PopulatesCacheWithDatabaseValues

willdot: Oh that's how you do it. I was trying to get it to return something from the query method I was calling. *Facepalm*

PR #244, opened 2019-05-17
Feature/update tasklist options

JB12298:


willdot:
I think it looks fine but merge conflict and failing build on Jenkins, although I can't see an error on Jenkins....

PR #255, opened 2019-05-21
Feature/edit return

JohnPadgettCivica:
1.I can see an "Edit" link in the table against all returns that I am able to resubmit. The rules are:

1.the return has been previously submitted (i.e. the status is "submitted")

2.there is no other resubmission started for that year and quarter (there can only be one "not submitted" record at any one time)
2.When I click on the edit link for a return: I am taken to the tasklist for that return with the data displayed
3.A copy of the record is made as soon as the user has clicked on the edit link
4.On the manage AATF returns screen the "previous" return record and the new record are both visible in the table. Refer to 2018 Q2 for example in prototye. Next to the previous record there are no options to continue or edit. Next to the new record I can click on a link to "continue" my resubmission.
5.Capture the full data set - don't overwrite the existing record.
6.Link the "new" record to its parent
7.Validate the URL so that users can't go directly to the URL
8.Note: for now ignore the expand/contract and show all rows in same font

Issues with testing against Entity framework meant it was easier creating integration tests for the copy handler rather than unit tests


src/EA.Weee.Web/Areas/AatfReturn/Controllers/ReturnsController.cs
```diff
@@ -13,6 +13,7 @@
     using Weee.Requests.AatfReturn;
     using Weee.Requests.Organisations;
 
+    //[RoutePrefix("Users")]
```
willdot: Has this been left in by mistake or just as documentation?

JohnPadgettCivica: Good spot... removed

PR #278, opened 2019-05-28
Add ability to filter the list of AATFs

stevospinks:
PBI: 71282


src/EA.Weee.Web/Areas/Admin/Views/Aatf/ManageAatfs.cshtml
```diff
@@ -16,15 +14,60 @@
         })
     </script>
 }
+
+<style>
```
JohnPadgettCivica: Can we add this to the weee-page.scss :)

stevospinks: Moved

PR #282, opened 2019-05-29
Bugs/create org for aatf bugs

willdot:
Missing bugs for Mansi


willdot:
> Can we just add unit test to ensure the breadcrumb it set

There are tests at the bottom of AddAatfControllerTests (When I looked at the commit, it doesn't load them as "Large diffs are not rendered by default")

JohnPadgettCivica:
> > Can we just add unit test to ensure the breadcrumb it set
> 
> There are tests at the bottom of AddAatfControllerTests (When I looked at the commit, it doesn't load them as "Large diffs are not rendered by default")

Just me being lazy then, doh

JohnPadgettCivica:
Think just needs a merge from develop...

willdot:
> Think just needs a merge from develop...

Ha I glanced over it as well at first. Merge with dev done. Wanna just check the change and then merge please :)

src/EA.Weee.Web/Areas/Admin/Controllers/AatfController.cs
```diff
@@ -87,7 +83,11 @@ public async Task<ActionResult> ManageAatfs(ManageAatfsViewModel viewModel)
             {
                 using (var client = apiClient())
                 {
-                    viewModel.AatfDataList = await GetAatfs(viewModel.Filter);
+                    viewModel = new ManageAatfsViewModel
+                    {
+                        AatfDataList = await GetAatfs(viewModel.Filter),
```
JohnPadgettCivica: Dunno if its me being blind again but cant see a test that sets the aatfDataList property of the viewmodel

willdot: Um test ManageAatfsPost_ModelError_GetAatfsMustBeRun does a check that the call to get aatfs is called.

JohnPadgettCivica: I think we are ok, seems covered elsewhere... one of those days it seems

PR #298, opened 2019-05-31
Feature/sprint10 content updates

SamGoodeySFW:


src/EA.Weee.Email/WeeeEmailService.cs
```diff
@@ -65,12 +65,13 @@ public async Task<bool> SendPasswordResetRequest(string emailAddress, string pas
             }
         }
 
-        public async Task<bool> SendOrganisationUserRequest(string emailAddress, string organisationName)
+        public async Task<bool> SendOrganisationUserRequest(string emailAddress, string organisationName, string userName)
```
JohnPadgettCivica: As there are no tests, can you create a basic test that validates the sendAsync method is called with a message generated from the previous steps

PR #307, opened 2019-06-04
Add aatf returns enabled check to returns controller

JohnPadgettCivica:
Issue that user could access the returns screen when aatf returns has been turned off


src/EA.Weee.Web.Tests.Unit/Areas/AatfReturn/Controller/ReturnsControllerTests.cs
```diff
@@ -41,6 +41,11 @@ public void ReturnsControllerInheritsAatfReturnBaseController()
             typeof(ReturnsController).BaseType.Name.Should().Be(typeof(AatfReturnBaseController).Name);
         }
 
+        [Fact]
+        public void ReturnsController_ShouldHaveValidateOrganisationActionFilterAttribute()
+        {
```
JB12298: Empty unit test?

JB12298:
The build is failing because of the index method in the AE controller.  I've fixed this on my branch so once the nil returns PR is merged, this shouldn't be an issue.

src/EA.Weee.Web.Tests.Unit/Areas/AatfReturn/Controller/ReturnsControllerTests.cs
JohnPadgettCivica: Quite clearing losing the plot... added unit test

PR #310, opened 2019-06-04
Feature/sprint10 bugs

JB12298:


src/EA.Weee.Web/Areas/Admin/Views/Aatf/Details.cshtml
```diff
@@ -176,11 +176,85 @@
                     <th class="govuk-table__header" scope="col">Email address</th>
                     <td class="govuk-table__cell">@Model.Organisation.BusinessAddress.Email</td>
                 </tr>
+                @if (Model.AssociatedAatfs.Any() || Model.AssociatedAEs.Any() || Model.AssociatedSchemes.Any())
```
JohnPadgettCivica: Nicer to create bool property on Model for example AnyRelatedEntities that would return true if any of these properties have values, then can unit test :)

JB12298: Added :)

PR #311, opened 2019-06-04
Add new AE to new/existing organisation

stevospinks:
PBI: 71748


src/EA.Weee.RequestHandlers.Tests.Unit/Admin/AddAatfRequestHandlerTests.cs
```diff
@@ -73,6 +71,7 @@ public async Task HandleAsync_ValidInput_AddsAatf()
             IWeeeAuthorization authorization = A.Fake<IWeeeAuthorization>();
 
             AatfData aatf = new AatfData(Guid.NewGuid(), "name", "approval number", (Int16)2019, A.Dummy<Core.Shared.UKCompetentAuthorityData>(), Core.AatfReturn.AatfStatus.Approved, A.Dummy<AatfAddressData>(), Core.AatfReturn.AatfSize.Large, DateTime.Now);
+            aatf.FacilityType = Core.AatfReturn.FacilityType.Ae;
```
JohnPadgettCivica: As only 2 facility types can we do a Theory test with both enum types and run the test for both values

stevospinks: Added

JohnPadgettCivica:
Looks like it just needs a merge from develop and then can complete

PR #327, opened 2019-06-07
added required error message so that the case is AATF/ATF

willdot:
Bug: 69531

### DIFF
--- a/src/EA.Weee.Core/AatfReturn/AatfAddressData.cs
+++ b/src/EA.Weee.Core/AatfReturn/AatfAddressData.cs
@@ -6,7 +6,7 @@
 
     public class AatfAddressData : AddressData
     {
-        [Required]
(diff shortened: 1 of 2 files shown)


src/EA.Weee.Core/AatfReturn/AatfAddressData.cs
```diff
@@ -6,7 +6,7 @@
 
     public class AatfAddressData : AddressData
     {
-        [Required]
+        [Required(ErrorMessage = "Enter AATF/ATF site name")]
```
JohnPadgettCivica: Can we create unit tests for these attributes on this model as its an annoying one, and if we change them and test fails at least we should ask ourselves should we bit changing

willdot: OK done

PR #331, opened 2019-06-07
Feature/edit ae org details bug fixes

SamGoodeySFW:
Bug 71966:
Mansi found an issue where this didn't work for PCS, made it more flexible so it can work for PCS and Aatf


JohnPadgettCivica:
Looks like build issue somewhere.

PR #332, opened 2019-06-07
Removed successfull org created screen.

willdot:
Task: 72037

Removed the success org create screen

Org gets created with a status of 1

When AATF/AE is created then org gets updated to 2.


JohnPadgettCivica:
Failing test by the looks of it

willdot:
Unit tests fixed

PR #344, opened 2019-06-10
Bug/ae nil return

willdot:
Bugs found on 71840

However, for John, can you check the implementation of my last commit for the redirecting if the user tries to bypass the activity list when they have no AATFs/AEs. No unit tests yet, as not sure if it's the correct way you meant.


JohnPadgettCivica:
I think you can just add your check into the ValidateOrganisationActionFilterAttribute attribute... and add the additional unit test to that class

PR #356, opened 2019-06-13
Feature/org search results

willdot:
PBI: 72222

Currently not sure on how to display the address, so simply displaying address 1, town, country,


src/EA.Weee.Web/Areas/Admin/Views/AddAatf/SearchResults.cshtml
```diff
@@ -36,7 +36,7 @@
                     m => m.SelectedOrganisationId,
                     Model.Results,
                     r => r.OrganisationId,
-                    r => r.Name,
+                    r => r.Name  +" " + r.AddressString,
```
JohnPadgettCivica: Rather than doing this can with have NameWithAddress property on the search result, then we can test that property

JohnPadgettCivica:
Looks like a test is failing

PR #357, opened 2019-06-13
fixed wrapping. Also fixed link that hadn't been updated in previous …

willdot:
…merge/commit

Bug#: 72210


JohnPadgettCivica:
Merge conflict, think due to fix I also made... if you take my change should be good

PR #362, opened 2019-06-14
Feature/aatf record nil return

SumeN:
For Bug 72156 - This PR records the NilReturn flag and redirects to report options for AATF return


src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/SubmitReturnHandlerTests.cs
```diff
@@ -132,7 +123,7 @@ public async Task HandleAsync_GivenReturnThatsAlreadySubmitted_StatusUpadateMust
 
             await handler.HandleAsync(message);
 
-            A.CallTo(() => @return.UpdateSubmitted(userId.ToString())).MustNotHaveHappened();
+            A.CallTo(() => @return.UpdateSubmitted(userId.ToString(), false)).MustNotHaveHappened();
```
JohnPadgettCivica: Change the false to A<string>._ as we don't really mind what value that is

src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/SubmitReturnHandlerTests.cs
```diff
@@ -114,7 +105,7 @@ public async Task HandleAsync_GivenReturn_StatusShouldBeUpdated()
 
             await handler.HandleAsync(message);
 
-            A.CallTo(() => @return.UpdateSubmitted(userId.ToString())).MustHaveHappened(Repeated.Exactly.Once)
+            A.CallTo(() => @return.UpdateSubmitted(userId.ToString(), false)).MustHaveHappened(Repeated.Exactly.Once)
```
JohnPadgettCivica: Can we change the test so that it is a [Theory] test and passes true & false to the test (i.e. for both nil return values) and then check that the UpdateSubmitted is called correctly with the value

src/EA.Weee.RequestHandlers/Mappings/ReturnMap.cs
```diff
@@ -42,7 +42,8 @@ public ReturnData Map(ReturnQuarterWindow source)
                 SubmittedBy = source.Return.SubmittedBy?.FullName,
                 SubmittedDate = source.Return.SubmittedDate,
                 ReturnReportOns = source.ReturnReportOns.Select(r => new ReturnReportOn(r.ReportOnQuestionId, r.ReturnId)).ToList(),
-                ReturnStatus = mapper.Map<ReturnStatus>(source.Return.ReturnStatus)
+                ReturnStatus = mapper.Map<ReturnStatus>(source.Return.ReturnStatus),
```
JohnPadgettCivica: Add ReturnMap tests to check check for the setting of the nil return bool

SumeN: Added [Theory] test

src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/SubmitReturnHandlerTests.cs
SumeN: Updated test

src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/SubmitReturnHandlerTests.cs
SumeN: Added [Theory] test

src/EA.Weee.Web/Areas/AatfReturn/Controllers/SelectReportOptionsNilController.cs
```diff
@@ -66,7 +66,7 @@ public virtual async Task<ActionResult> Index(SelectReportOptionsNilViewModel vi
             {
                 var request = requestCreator.ViewModelToRequest(viewModel);
                 await client.SendAsync(User.GetAccessToken(), request);
-                await client.SendAsync(User.GetAccessToken(), new SubmitReturn(viewModel.ReturnId));
+                await client.SendAsync(User.GetAccessToken(), new SubmitReturn(viewModel.ReturnId, true));
```
JohnPadgettCivica: Update the unit test to ensure the request is called with the value of true

PR #376, opened 2019-06-18
Feature/add back return links external

SumeN:
For PBI 72221 - Add "Back" and "Back to check return" links. There are couple of screens (given in task 72418) for which the links are also implemented

### DIFF
--- a/src/EA.Weee.Web/Areas/AatfReturn/Controllers/ReusedOffSiteController.cs
+++ b/src/EA.Weee.Web/Areas/AatfReturn/Controllers/ReusedOffSiteController.cs
@@ -39,13 +39,6 @@
 
             await SetBreadcrumb(organisationId, BreadCrumbConstant.AatfReturn);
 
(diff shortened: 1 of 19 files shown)


src/EA.Weee.Web/Areas/AatfReturn/Controllers/ReusedOffSiteController.cs
```diff
@@ -39,6 +39,13 @@ public virtual async Task<ActionResult> Index(Guid organisationId, Guid returnId
 
             await SetBreadcrumb(organisationId, BreadCrumbConstant.AatfReturn);
 
+            string path = (Request != null && Request.UrlReferrer != null) ? Request.UrlReferrer.AbsolutePath : String.Empty;
```
JohnPadgettCivica: I wouldn't worry doing this unless there is specific acceptance criteria... if they click back after selecting yes we should be ok with blank selection

SumeN: ok

PR #384, opened 2019-06-24
Feature/delete aatf

SumeN:
For PBI - 72241.  This PBI covers delete of AATF and AE


src/EA.Weee.Web/Areas/Admin/Controllers/AatfController.cs
```diff
@@ -225,6 +227,43 @@ public async Task<ActionResult> ManageContactDetails(AatfEditContactAddressViewM
             return View(viewModel);
         }
 
+        [HttpGet]
+        public async Task<ActionResult> Delete(Guid id, Guid organisationId, FacilityType facilityType)
```
JohnPadgettCivica: Can we add the  [AuthorizeInternalClaims(Claims.InternalAdmin)] attribute to the action and create unit test

Maybe also add to the ManageContact details and ManageAatf details actions

SumeN: Added

PR #386, opened 2019-06-24
Feature/email when user joins org

SamGoodeySFW:


src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -425,11 +425,21 @@ public async Task<ViewResult> JoinOrganisation(Guid organisationId)
...
+                    {
+                        anyActiveUsers = false;
+                    }
+
                     var model = new JoinOrganisationViewModel
                     {
                         OrganisationId = organisationId,
-                        OrganisationName = organisationData.DisplayName
+                        OrganisationName = organisationData.DisplayName,
+                        AnyActiveUsers = anyActiveUsers
```
JohnPadgettCivica: Set the AnyActiveUsers property to activeUsers.Any() then you don't need the if

PR #397, opened 2019-06-27
Order PCS by scheme name

SumeN:
For PBI - 72572


src/EA.Weee.RequestHandlers/Scheme/GetSchemesExternalHandler.cs
```diff
@@ -34,7 +34,7 @@ public async Task<List<SchemeData>> HandleAsync(GetSchemesExternal message)
 
             return schemes.Where(s => (s.SchemeStatus == SchemeStatus.Approved) || (s.SchemeStatus == SchemeStatus.Withdrawn))
                 .Select(s => schemeMap.Map(s))
-                .OrderBy(sd => sd.Name)
+                .OrderBy(sd => sd.SchemeName)
```
JohnPadgettCivica: Can we write a unit tests to check the order :)

PR #405, opened 2019-07-01
Remove PCS for selected return

SumeN:
For bug 72662 - Delete PCS

### DIFF
--- a/src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/RemoveReturnSchemeHandlerTests.cs
+++ b/src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/RemoveReturnSchemeHandlerTests.cs
@@ -52,7 +52,7 @@
 
             await handler.HandleAsync(request);
 
-            A.CallTo(() => returnSchemeDataAccess.RemoveReturnScheme(schemeIds)).MustHaveHappened(Repeated.Exactly.Once);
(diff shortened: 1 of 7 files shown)


src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/RemoveReturnSchemeHandlerTests.cs
```diff
@@ -52,7 +52,7 @@ public async Task HandleAsync_GivenSchemeId_ReturnSchemeDeleted()
 
             await handler.HandleAsync(request);
 
-            A.CallTo(() => returnSchemeDataAccess.RemoveReturnScheme(schemeIds)).MustHaveHappened(Repeated.Exactly.Once);
+            A.CallTo(() => returnSchemeDataAccess.RemoveReturnScheme(schemeIds, A<Guid>._)).MustHaveHappened(Repeated.Exactly.Once);
```
JohnPadgettCivica: Can we check the return Id here also?

src/EA.Weee.RequestHandlers/AatfReturn/ReturnSchemeDataAccess.cs
```diff
@@ -39,11 +39,11 @@ public async Task<Organisation> GetOrganisationByReturnId(Guid returnId)
             return @return?.Organisation;
         }
 
-        public async Task RemoveReturnScheme(List<Guid> schemeIds)
+        public async Task RemoveReturnScheme(List<Guid> schemeIds, Guid returnId)
```
JohnPadgettCivica: Can we write an integration test for this method to ensure its only the returns we won't data deleted

PR #408, opened 2019-07-02
Feature/by pass url bug

SumeN:
For bug 72687 - By passing URL and added the pending tests for delete PR.

### DIFF
--- a/src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/ReturnSchemeDataAccessTests.cs
+++ b/src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/ReturnSchemeDataAccessTests.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Data.Entity;
     using System.Linq;
     using System.Threading.Tasks;
     using DataAccess;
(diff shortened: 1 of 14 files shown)


src/EA.Weee.RequestHandlers.Tests.Unit/AatfReturn/ReturnSchemeDataAccessTests.cs
```diff
@@ -80,5 +80,48 @@ public async Task GetOperatorByReturnId_GivenReturnIdDoesNotExist_NullShouldBeRe
...
+            A.CallTo(() => weeeReceivedAmount.Id).Returns(A.Dummy<Guid>());
+            list.Add(weeeReceivedAmount);
+            A.CallTo(() => context.WeeeReceivedAmount).Returns(dbHelper.GetAsyncEnabledDbSet(list));
+
+            A.CallTo(() => context.WeeeReceived).Returns(dbHelper.GetAsyncEnabledDbSet(weeeReceived));
+            A.CallTo(() => context.ReturnScheme).Returns(dbHelper.GetAsyncEnabledDbSet(new List<ReturnScheme>() { returnSchemeNoMatch, returnSchemeMatch }));
+
+            await dataAccess.RemoveReturnScheme(schemeIds, returnId);
+
+            A.CallTo(() => context.WeeeReceived.Remove(weee)).MustHaveHappened(Repeated.Exactly.Once);
```
JohnPadgettCivica: Can we also check that the return that does not match is still in the context

SumeN: Added

PR #409, opened 2019-07-02
added link to privacy policy and checkbox

willdot:
PBI: 72582

### DIFF
--- a/src/EA.Weee.Web/ViewModels/NewUser/UserCreationViewModel.cs
+++ b/src/EA.Weee.Web/ViewModels/NewUser/UserCreationViewModel.cs
@@ -40,5 +40,8 @@
 
         [MustBeTrue(ErrorMessage = "Confirm that you've read and accepted the terms and conditions")]
         public bool TermsAndConditions { get; set; }
+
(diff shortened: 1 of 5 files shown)


src/EA.Weee.Web/ViewModels/NewUser/UserCreationViewModel.cs
```diff
@@ -40,5 +40,8 @@ public class UserCreationViewModel
 
         [MustBeTrue(ErrorMessage = "Confirm that you've read and accepted the terms and conditions")]
         public bool TermsAndConditions { get; set; }
+
+        [MustBeTrue(ErrorMessage = "Confirm that you've read and accepted the privacy policy")]
+        public bool PrivacyPolicy { get; set; }
```
JohnPadgettCivica: Can we add unit test to the model for this attribute - and maybe the T & C one also

JohnPadgettCivica:
Just on a call now and it seems the privacy notice text should be in a progressive disclosure

willdot:
Ok, changed it to use the progressive disclosure view that I found.

PR #422, opened 2019-07-04
Feature/check return change links for options

SamGoodeySFW:
[PBI: 72573](https://team.sfwltd.co.uk:8081/tfs/Projects/EA_PRSD/_workitems#_a=edit&id=72573)


JohnPadgettCivica:
This has reverted some of the changes I made to the column styles, I updated to 1/3 rather than 1/4...

PR #435, opened 2019-07-05
Feature/google tags

willdot:
PBI: 72796

There is an attachment that I've added that is a duplicate of the original file but with a done or not column

Note there are a few not dones, as these links don't exist yet.


src/EA.Weee.Web/Areas/AatfReturn/Views/CheckYourReturn/Index.cshtml
```diff
@@ -285,25 +310,30 @@
...
+                        </p>
+                    }
                 </div>
 
                 <div class="govuk-grid-column-one-third weee-tonnage-right-aligned-display">
                     <p>
                         @if (Model.ReportOnDisplayOptions.DisplayNonObligated)
                         {
-                            @Html.NavigationRouteLink("Change", "Change non obligated", AatfRedirect.NonObligatedRouteName, new { returnId = Model.ReturnId, dcf = false })
+                            @(this.WeeeGds().ActionLinkWithEventTracking("Change",
```
JohnPadgettCivica: The NavigationRouteLink html adds a hidden span with some explanation text for screenreaders, i.e. "Change non obligated" just loaded this up on and the hidden text is This link opens in a new browser window... so isn't correct

JohnPadgettCivica:
Can you close this PR and create against release/release-aatf-external and also merge from release/release-aatf-external as there as been a change to update Add data / edit data to add data / edit data

PR #447, opened 2019-07-09
Restrict when submissions can be continued

stevospinks:
Hides the link and redirects to an error page when the URL is accessed directly and the reporting window has closed.


JohnPadgettCivica:
Made a comment on Wills PR around doing the checks in the controller attributes, suspect relevant here

stevospinks:
Will has made the changes in his branch.
As his is based off mine, I don't think it makes sense for me to change it here too

PR #459, opened 2019-07-10
Update the search to return organisations that have a rejected PCSs but some AATFs / AEs

stevospinks:
It turned out my initially complex LINQ query could be simplified quite a bit...

PBI: 72952

### DIFF
--- a/src/EA.Weee.RequestHandlers.Tests.DataAccess/Search/FetchOrganisationSearchResultsForCache/FetchOrganisationSearchResultsForCacheDataAccessTests.cs
+++ b/src/EA.Weee.RequestHandlers.Tests.DataAccess/Search/FetchOrganisationSearchResultsForCache/FetchOrganisationSearchResultsForCacheDataAccessTests.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
+    using AutoFixture;
     using EA.Weee.Core.Search;
     using EA.Weee.DataAccess;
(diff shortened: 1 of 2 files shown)


src/EA.Weee.RequestHandlers.Tests.DataAccess/Search/FetchOrganisationSearchResultsForCache/FetchOrganisationSearchResultsForCacheDataAccessTests.cs
```diff
@@ -275,42 +278,68 @@ public async Task FetchCompleteOrganisations_WithSeveralResults_ReturnsResultsOr
...
         }
 
-        /// <summary>
-        /// Ensures that search results representing organisations which are partnerships will use
-        /// the 'TradingName' column from the database as the organisation name.
-        /// </summary>
-        /// <returns></returns>
         [Fact]
-        public async Task FetchCompleteOrganisations_OrganisationHasNoScheme_OrganisationReturned()
+        public async Task FetchCompleteOrganisations_OrganisationHasNoScheme_OrganisationNotReturned()
```
JohnPadgettCivica: This looks like it tests where the Organisation has no related entities, I think :) If so can we rename otherwise all looks good

stevospinks: Good point, renamed

PR #475, opened 2019-07-18
Report: UK WEEE Received at AATFs

stevospinks:
PBI: 73020


src/EA.Weee.RequestHandlers/Admin/Reports/GetUkWeeeAtAatfsCsv/GetUkWeeeAtAatfsCsvDataAccess.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        /// <summary>
+        /// Returns all data returns for the specified compliance year.
+        /// Each data return will be pre-loaded with the current version's
+        /// set of collected amounts and delivered amounts.
+        /// The results will not be ordered.
+        /// </summary>
+        public async Task<IEnumerable<PartialAatfReturn>> FetchPartialAatfReturnsForComplianceYearAsync(int complianceYear)
+        {
+            var parentIds = await context.Returns
```
JohnPadgettCivica: Is the checking of the parent Id's to try and calculate the latest submitted returns?

stevospinks: Yes, it's essentially saying 'if an AATF Return is a parent, then ignore it'

JohnPadgettCivica: The parent Id is just to store which return a return was copied from, i.e. when the edit link is clicked
I think this query will probably be excluding the latest as each edit will have a parent, and each edit will be the most recent? (I think) Probably need to group by Organisation and take the top one ordered by Submitted date desc and where return status = 2

stevospinks: I think it will still work, as the latest edit will not be a parent to any other report, so it shouldn't be excluded.
The way you suggest sounds more logical though, so I'll implement that - the code will be easier to understand too!

stevospinks: Pushed changes and test updates

JohnPadgettCivica:
Looks good, just need to do a fresh merge from develop as some conflicts...

PR #501, opened 2019-08-01
Feature/alter summary aatf returns report

JohnPadgettCivica:
Need to add additional unit tests for the stored procedure but if you can start to do a review then I can get the testers testing while I finish off unit test


SamGoodeySFW:
Looks good to me so far

SumeN:
In the Stored Proc, 
- line 207 (AND r.Id NOT IN (SELECT ReturnId FROM @RETURN)) is not needed as there isn't any records in @return
- i think the started and not started records would either show as resubmission or first submission

PR #517, opened 2019-08-20
Feature/add aatf for cy integration tests

SumeN:
For bugs  in 73426. Removed new handlers and used the add handler/requests for Copy.


src/EA.Weee.RequestHandlers/AatfReturn/AatfTaskList/FetchAatfDataAccess.cs
```diff
@@ -38,9 +38,16 @@ public async Task<List<Aatf>> FetchAatfByReturnId(Guid returnId)
             return await context.ReturnAatfs.Where(r => r.Return.Id.Equals(returnId)).Select(r => r.Aatf).ToListAsync();
         }
 
-        public async Task<Aatf> FetchByApprovalNumber(string approvalNumber)
+        public async Task<Aatf> FetchByApprovalNumber(string approvalNumber, int? year = null)
```
JohnPadgettCivica: We should always supply a year so doesnt need to be nullable

src/EA.Weee.RequestHandlers/Admin/AddAatfRequestHandler.cs
```diff
@@ -66,7 +66,28 @@ public async Task<bool> HandleAsync(AddAatf message)
                 panArea = await commonDataAccess.FetchLookup<PanArea>(message.Aatf.PanAreaData.Id);
             }
 
-            var aatf = new Domain.AatfReturn.Aatf(
+            if (message.AatfId == null)
```
JohnPadgettCivica: Rather than doing this here is it possible to just pass the AATF id in the constructor and if its null then set it

JohnPadgettCivica:
Does this change also contain the changes to the validator to ensure that the aatf doesnt already have an entry for the year?

src/EA.Weee.RequestHandlers/Admin/AddAatfRequestHandler.cs
SumeN: Done

PR #531, opened 2019-08-27
Bugfix/non obligated report

JohnPadgettCivica:
Task 73832:BUG: Downloaded report has all the column irrespective of selected reporting options for non-obligated.
Task 73833:BUG: "Cancel" link is displayed beside Continue button on copy-paste screen on Non-obligated screen
Task 70694:BUG: AATF/AE tab layout is not same as internal PCS details screen


src/EA.Weee.Web/Scripts/weee-application.js
```diff
@@ -109,8 +109,12 @@
         locale: "en"
 	    }
     );
+
+    $('.govuk-tabs__tab--selected').focus();
 });
```
SumeN: Good one. i think the only downside is even when the page loads it will get highlighted

JohnPadgettCivica: Which I think it should... looks like AATF tab that is initially loaded doesnt get the selected style... think ill add that

PR #537, opened 2019-08-28
Bug/aatf ae details report

willdot:
BUGS:
73884
73898

### DIFF
--- a/src/EA.Weee.Web/Areas/Admin/Controllers/AatfReportsController.cs
+++ b/src/EA.Weee.Web/Areas/Admin/Controllers/AatfReportsController.cs
@@ -425,7 +425,7 @@
         {
             using (var client = apiClient())
             {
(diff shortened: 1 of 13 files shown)


src/EA.Weee.Web/Areas/Admin/Controllers/AatfReportsController.cs
```diff
@@ -425,7 +425,7 @@ private async Task PopulateFilters(AatfAeDetailsViewModel model)
         {
             using (var client = apiClient())
             {
-                model.ComplianceYears = new SelectList(FetchAllAatfComplianceYears());
+                model.ComplianceYears = new SelectList(new List<int> { 2019, 2020, 2021 });
```
JohnPadgettCivica: Think we would need a new request handler for this... should query all the unique Compliance years in the AATF table

PR #539, opened 2019-08-28
Feature/aatf ae details report change

willdot:
Changes to PBI 73690 (the A/C in bold)

Task 73911


JohnPadgettCivica:
Just need to update the 20190827-1654-create-sp-getAatfAeDetailsCsvData.sql  script date as this has already been executed against the databases so won't be updated

PR #540, opened 2019-08-28
Feature/prevent duplication of entities

SamGoodeySFW:
[PBI 73881](https://team.sfwltd.co.uk:8081/tfs/Projects/EA_PRSD/_workitems#_a=edit&id=73881)


JohnPadgettCivica:
The change I think should only show the releated AATFs and AE's for the year selected in the drop down, do these changes do this?

SamGoodeySFW:
> The change I think should only show the releated AATFs and AE's for the year selected in the drop down, do these changes do this?

It doesn't, it shows the most recent AATFs and AEs regardless of the drop down selection, should be linking what is returned by the compliance year selected?

PR #546, opened 2019-08-29
Feature/manage aatf show obligated columns

SamGoodeySFW:
PBI 73905


JohnPadgettCivica:
Also need to update the weee-returns-aatf-mobile.scss and weee-returns-ae-mobile.scss files as there are more columns now

PR #576, opened 2019-09-16
Feature/copy pcs from previous quarter

willdot:
PBI: 74211

For the deselect pcs option, I've opted to add a button, because to cope with no javascript, there would need to be a button anyway

### DIFF
--- a/src/EA.Weee.RequestHandlers/AatfReturn/GetPreviousQuarterSchemesHandler.cs
+++ b/src/EA.Weee.RequestHandlers/AatfReturn/GetPreviousQuarterSchemesHandler.cs
@@ -2,6 +2,7 @@
 {
     using EA.Prsd.Core.Mediator;
     using EA.Weee.Domain.AatfReturn;
+    using EA.Weee.Domain.DataReturns;
     using EA.Weee.RequestHandlers.Security;
     using EA.Weee.Requests.AatfReturn;
     using System;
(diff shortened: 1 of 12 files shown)


src/EA.Weee.RequestHandlers/AatfReturn/GetPreviousQuarterSchemesHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (returnSchemes == null)
+            {
+                return new PreviousQuarterReturnResult();
+            }
+
+            List<ReturnScheme> lastSchemes = new List<ReturnScheme>();
+
+            // Grab the last submitted return
+            Return lastReturn = allReturns.FirstOrDefault(p => p.Quarter == lastQ);
+            lastSchemes.AddRange(returnSchemes.Where(p => p.ReturnId == lastReturn.Id));
```
JohnPadgettCivica: last return could be NULL if lastQ is null - i.e. there are no previous returns

src/EA.Weee.Web/Areas/AatfReturn/Controllers/SelectYourPCSController.cs
```diff
@@ -37,7 +37,7 @@ public SelectYourPcsController(Func<IWeeeClient> apiclient, BreadcrumbService br
         }
 
         [HttpGet]
-        public virtual async Task<ActionResult> Index(Guid organisationId, Guid returnId, bool reselect = false)
+        public virtual async Task<ActionResult> Index(Guid organisationId, Guid returnId, bool reselect = false, bool copyPrevious = false, bool clearSelections = false)
         {
             if (reselect)
             {
```
JohnPadgettCivica: Think we need to refactor this a little can we move the code from the Reselect method into the main Index action as we will need to do the copy functionality if its a reselection, currently there is an exception if you come from the Change your pcs selection link on the recieved pcs list

So set the SelectedSchemes  and reselect properties in here.

src/EA.Weee.RequestHandlers/AatfReturn/GetPreviousQuarterSchemesHandler.cs
willdot: I've already done a check to see if all returns for that organisation id, that aren't the current return id and is submitted. Lines 34. That should work?

PR #588, opened 2019-09-19
Feature/aatf filters

SumeN:
For pbi - 74219


src/EA.Weee.Web/Areas/Admin/ViewModels/Aatf/FilteringViewModel.cs
```diff
@@ -11,5 +16,44 @@ public class FilteringViewModel
...
+
+        public bool SelectApproved { get; set; }
+
+        public bool SelectCancelled { get; set; }
+
+        public bool SelectSuspended { get; set; }
+
+        public List<UKCompetentAuthorityData> CompetentAuthorityOptions { get; set; }
+
+        public List<int> SelectedStatus
```
JohnPadgettCivica: Can we add tests for this property

SumeN: added

PR #609, opened 2022-03-04
Feature/nuget upgrade

JohnPadgettCivica:


JohnPadgettCivica:
Completed nuget and .NET upgrade along with Integration test framework

PR #625, opened 2022-03-21
AATF Evidence Choose Site

NateHogan:
Current issue with routing

### DIFF
--- a/src/EA.Weee.Web/Areas/AatfEvidence/AatfEvidenceAreaRegistration.cs
+++ b/src/EA.Weee.Web/Areas/AatfEvidence/AatfEvidenceAreaRegistration.cs
@@ -21,7 +21,7 @@
 
             context.MapLowercaseDashedRoute(
                 name: AatfEvidenceRedirect.Default,
-                url: "aatf-evidence/{aatfId}/choose-site/",
+                url: "aatf-evidence/{organisationId}/choose-site/",
                 defaults: new { action = "Index", controller = "SelectYourAatf" },
                 namespaces: new[] { typeof(SelectYourAatfController).Namespace });
         }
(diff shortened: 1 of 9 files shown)


JohnPadgettCivica (changes requested):
For the Index Post could we create a test that checks the setting of the breadcrumb similar to the Get test.

Index Get – test that the correct model is returned.

Index Post – test that when the model is invalid that the api and mapper are re-called and that the view model is returned. Perhaps move the api and mapper to its own method and call from the GET and POST as they look pretty similar.

I’m not sure we need to do ModelState.RunCustomValidation(model); at all. Try without.

The AatfDataToHomeViewModelMap and AatfDataToSelectYourAatfViewModelMap classes are pretty much identitical.
I would suggest… pulling the below code into its own class, call it AatfFilter, which could have single method Filter, for example.

Public class AatfFilter : IFilter{

	Public List<AatfData> Filter(List<AatfData> items, FacilityType facility){
		var filteredAatfList = RemoveOlderAatfs(items);

            var selectedAatfsOrAes = new List<AatfData>();

            selectedAatfsOrAes.AddRange(filteredAatfList.Where(m => m.FacilityType == facility));

            foreach (var aatf in selectedAatfsOrAes)
            {
                aatf.AatfContactDetailsName = aatf.Name + " (" + aatf.ApprovalNumber + ") - " + aatf.AatfStatus;
            }

		return selectedAatfsOrAes;
}
}

This code can then be injected using the IFilter interface into AatfDataToHomeViewModelMap and AatfDataToSelectYourAatfViewModelMap and we could do,

Var items = aatfFilter.Filter(source.AatfList, source.FacilityType);

This would mean pulling some of the unit tests into their own test class.

If none of this makes sense we can chat in the morning 


Unit tests for the SelectYourAatfViewModel model to check that it has the attributes expected. 

SelectYourAatfViewModel we don’t need to the FacilityType property as it will always be an AATF.

PR #638, opened 2022-03-28
Test Area - Organisation Search

NateHogan:
Removes external user paragraph from view org search view

Adds new request/request handler for partial searching organisations
Adds unit tests


GuidoWorkDev (commented):
If you check acceptance criteria - point #2 - it says: "content states that user must be logged on as an internal user and can search across all organisations" - so I think you should amend the paragraph rather than delete it.  this comment concerns the 2 cshtml files.
Everything else is looking good.

PR #650, opened 2022-03-31
Feature/aatf view list of draft notes ext

JohnPadgettCivica:


JohnPadgettCivica:
I've added a pull request for what we should so for the model mapper in the action. 

I think there is a way that we could make things work with a async action so we don't have to do .Result on the API call. I'll try and see if it works if you want to write integration tests

src/EA.Weee.Core/AatfEvidence/EvidenceNoteData.cs
```diff
@@ -6,19 +6,43 @@
     using System.Linq;
     using System.Text;
     using System.Threading.Tasks;
+    using AatfReturn;
+    using Prsd.Core;
 
     public class EvidenceNoteData
     {
         //TODO: Other evidence note properties required
         public int Reference { get; set; }
```
JohnPadgettCivica, line 15: Can we return the Note Id primary key on this (and map it) we will have to use it to generate the View / Edit links

JohnPadgettCivica:
I think if you are happy with the changes I made in https://github.com/DEFRA/prsd-weee/pull/655 we merge that in to yours and then we can get your branch into test for Sneha. Then you can have a look at writing the tests.

Tests I think we need,

Integration test for GetAatfNotesRequestHandler have a look at the other tests in EA.Weee.Integration.Tests.
Unit tests for GetAatfNotesRequestHandler
Unit tests for EditDraftReturnedNotesMapper
Unit tests for EditDraftReturnNotesViewModelMap that I added in the other branch
Simple unit test for EditDraftReturnNotesViewModelMap

PR #660, opened 2022-04-05
left-aligned global title "Waste Electrical and Electronic Equipment …

GuidoWorkDev:
…(WEEE)" in layout GovUK template
I have shrunk the logo to the left to enable the title to move to the left.

### DIFF
--- a/src/EA.Weee.Web/Views/Shared/_Layout.GovUK.cshtml
+++ b/src/EA.Weee.Web/Views/Shared/_Layout.GovUK.cshtml
@@ -50,7 +50,7 @@
     <header class="govuk-header @Html.Raw(ViewBag.HeaderClass ?? string.Empty)" role="banner" data-module="govuk-header">
         <div class="govuk-header__container govuk-width-container">
             <div class="header-global">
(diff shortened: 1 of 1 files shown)


JohnPadgettCivica (changes requested):
I think for this one we need to take a closer look at how the header https://design-system.service.gov.uk/components/header/ works and what is different to the WEEE one.

src/EA.Weee.Web/Views/Shared/_Layout.GovUK.cshtml
```diff
@@ -50,7 +50,7 @@
     <header class="govuk-header @Html.Raw(ViewBag.HeaderClass ?? string.Empty)" role="banner" data-module="govuk-header">
         <div class="govuk-header__container govuk-width-container">
             <div class="header-global">
-                <div class="govuk-header__logo">
+                <div class="govuk-header__logo" style="width: 200px;">
```
meglenapnovoselska, line 53: have a look at the _index.css, there is a padding set in the class govuk-header__logo

also that logo is used in Maintenance.cshtml file. Might need a change there too.

PR #663, opened 2022-04-06
reduced the size of the H1 title to fit more text on first line.

GuidoWorkDev:


JohnPadgettCivica (changes requested):
Have a look at how the EditDraftReturnedNotesOverview.cshtml goes full width. I suggest we do something similar to that on the Choose site screen.

Even nice is in _Layout.GovUk.Sections.chsml if we do an additional If statement, something like
@if (IsSectionDefined("main_content-full-width"))
        {
            <section id="main-content">
    <div class="govuk-grid-row">
        <div class="govuk-grid-column-full">
                    @RenderSection("main_content-full-width", required: false)
</div>
</div>
</section>
           
And then we can define the sections in the page like

@section main_content_full_width
{

}

PR #705, opened 2022-04-22
CCN050 WEEE PECR application Cookie updates

martin-461:
CCN050 WEEE PECR application Cookie updates


src/EA.Weee.Web/Scripts/govuk/govuk-template.js
```diff
@@ -57,22 +57,21 @@
       }
     }
     return null;
-  };
-}).call(this);
```
JohnPadgettCivica, line 61: @sreedhar-civica  There is a problem with the way this is implemented. We often upgrade the GOV.UK design system which means this template.js will get updated. We shouldn't change these files directly we should have custom js that overrides when we need to override.

martin-461: @JohnPadgettCivica am happy to rewrite this though can I just double check something. If we leave the govuk-template.js as untouched would this still be relevant? Reason I mention is the functionality before in this file used to add a cookie when the main page was loaded (cookie=seen_cookie_message) which is now incorrect behaviour. The functionality has now changed a user has to accept or reject the cookie message and we have to act according. Just wanted to double check as I can put this back to how it was but its not really going to be correct for future development.

JohnPadgettCivica: @martin-461 Unfortunately it's unlikely to remain untouched. In the current project work we have upgraded the GOV.UK Design system and that file has already changed. Suspect we will also have to do further upgrades in the future...

martin-461: No worries just thought I'd check, shouldn't be too tricky to refactor :)

JohnPadgettCivica: :)

PR #708, opened 2022-04-22
Remove the submitted date from the note table and use the status hist…

JohnPadgettCivica:
Remove the submitted date and by from the note table and use the note status history


meglenapnovoselska:
is that still valid in EvidenceNoteRowViewModelMap class -  SubmittedBy = source.SubmittedDate.HasValue ? source.AatfData.Name : string.Empty. 
Does it need to be changed to ChangedBy from NoteHistory object?

meglenapnovoselska (approved):

JohnPadgettCivica:
> 

No, I may rename the submitted by field at some point as its a bit misleading. As its the AATF that submitted the note rather than the user.

PR #716, opened 2022-04-26
Submitted date now presented in GMT

NateHogan:


src/EA.Weee.Web/ViewModels/Shared/Mapping/ViewEvidenceNoteViewModelMap.cs
```diff
@@ -33,6 +33,7 @@ public ViewEvidenceNoteViewModel Map(ViewEvidenceNoteMapTransfer source)
                 Type = source.EvidenceNoteData.Type,
                 StartDate = source.EvidenceNoteData.StartDate,
                 EndDate = source.EvidenceNoteData.EndDate,
+                SubmittedDate = source.EvidenceNoteData.SubmittedDate.HasValue ? source.EvidenceNoteData.SubmittedDate.Value.ToString("dd/MM/yyyy HH:mm:ss \"(GMT)\"") : string.Empty,
```
meglenapnovoselska, line 36: shall we keep as a constant somewhere in a shared place? probably will be reused
also is there any unit tests to be added/modified?

meglenapnovoselska: would be nice to have a helper method which does all this the .HasValalue check and converts the string or returns an empty one.

JohnPadgettCivica: yeah agree with @meglenapnovoselska. Make it extension method in DisplayExtensions, which can process a nullable datetime. Also need unit test to be updated to ensure that the ViewEvidenceNoteViewModelMap sets the string correctly.

NateHogan: Yeah agreed and implemented, both a constant and an extension method has been added

PR #721, opened 2022-04-27
Feature/43751 update success message

GuidoWorkDev:
Modified the Set method - the mapper will write the status when called.


src/EA.Weee.Web/ViewModels/Shared/Mapping/ViewEvidenceNoteViewModelMap.cs
```diff
@@ -89,7 +89,7 @@ private void SetSuccessMessage(EvidenceNoteData note, object noteStatus, ViewEvi
                                 $"You have successfully saved the evidence note with reference ID E{note.Reference} as a draft";
                             break;
                         case NoteStatus.Approved:
-                            model.SuccessMessage = "success message TODO";
+                            model.SuccessMessage = $"You have successfully approved the evidence note with reference ID E{note.Reference}";
```
JohnPadgettCivica, line 92: update the unit test to check for this case.

JohnPadgettCivica:
You just need to merge in latest develop as there is a merge conflict in ViewEvidenceNoteViewModelMapTests.cs. You will need to take both changes,

PR #727, opened 2022-04-28
Add radio button with hints

JohnPadgettCivica:
Add hint text to radio button options


src/EA.Weee.Web/Areas/Scheme/ViewModels/ManageEvidenceNotes/ReviewEvidenceNoteViewModel.cs
```diff
@@ -22,5 +22,16 @@ public Core.AatfEvidence.NoteStatus SelectedEnumValue
         public ReviewEvidenceNoteViewModel() : base(new List<string> { "Approved" })
         {
         }
+
+        public Dictionary<string, string> HintItems
+        {
+            get
+            {
+                return new Dictionary<string, string>();
+                //{
```
JohnPadgettCivica, line 31: This is only here temporarily as an example

NateHogan (approved):
Looks good, I think there were some unused imports but it looks like with the temporary comment too that they could be removed during another PR

PR #767, opened 2022-05-13
Basic Obligations Controller

NateHogan:


JohnPadgettCivica:
Think a test is failing. While we are doing this can we add tests to test the controller base classes and the authorise attribute :)

PR #789, opened 2022-05-23
Cleaned up code and fix of couple of bugs - Reorder of the Recipient Name, Reorder of the Status filter, Remove the look ahead for 2 of the filters

meglenapnovoselska:
Change the order of the recipient filter to be alphabetical
Change the order of the Status filter as per UX designed request
Remove the with look ahead for the waste and status filter

### DIFF
--- a/src/EA.Weee.Web/Areas/Aatf/Controllers/ManageEvidenceNotesController.cs
+++ b/src/EA.Weee.Web/Areas/Aatf/Controllers/ManageEvidenceNotesController.cs
@@ -275,24 +275,14 @@
 
             var schemeData = resultAllNotes.Select(x => x.SchemeData)
                    .Distinct(new SchemeDataComparer())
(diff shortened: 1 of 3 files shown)


src/EA.Weee.Web/Areas/Aatf/Controllers/ManageEvidenceNotesController.cs
```diff
@@ -275,24 +275,14 @@ private async Task SetBreadcrumb(Guid organisationId, string activity)
 
             var schemeData = resultAllNotes.Select(x => x.SchemeData)
                    .Distinct(new SchemeDataComparer())
-                    .OrderByDescending(s => s.SchemeName).ToList();
+                    .OrderBy(s => s.SchemeName).ToList();
```
JohnPadgettCivica, line 278: Add a unit test for this when you get around to it

PR #803, opened 2022-05-25
Feature/45033 bug fix

GuidoWorkDev:
this is the fix for bug 45033 on its own - other bugs will follow in their own PR.


meglenapnovoselska (commented):
approve

PR #814, opened 2022-05-27
Fix the date picker next prev icons

JohnPadgettCivica:


JohnPadgettCivica:
Fix the jquery ui bundle

PR #820, opened 2022-05-31
PR- 45457 Clicking on the error message link is not shifting the focus to the element

meglenapnovoselska:
Fixing a bug which was causing clicking on the error message on top to not display the field where the error message came from.


src/EA.Weee.Web/Areas/Scheme/Views/TransferEvidence/TransferTonnage.cshtml
```diff
@@ -118,15 +118,15 @@
                                         <span class="govuk-visually-hidden" id="category-received-@category.CategoryId-@i.ToString()">Transfer received (tonnes) for @note.SubmittedBy and @category.CategoryDisplay</span>
                                         @Html.Gds().ValidationMessageFor(m => Model.TransferCategoryValues[elementCount].Received)
                                         @Html.TextBoxFor(m => Model.TransferCategoryValues[elementCount].Received,
-                                            new { @id = @elementCount + "Received" + @category.CategoryId, @class = "govuk-input tonnage-entry", aria_labelledby = "category-received-" + @category.CategoryId + "-" + @i.ToString(), autocomplete = "off", maxlength = CommonMaxFieldLengths.Tonnage })
```
JohnPadgettCivica, line 121: We either have to keep this id as it is as it is used to calculate the totals at the bottom of the screen or we need to update the javascript to calculate these totals. Or have some custom javascript for this screen that clicking on the link take us to the correct control.

PR #822, opened 2022-05-31
fix to resolve transfer notes being returned for a scheme

JohnPadgettCivica:


src/EA.Weee.DataAccess.Tests.Integration/EvidenceNote/NotesDataAccessIntegration.cs
```diff
@@ -64,6 +64,49 @@ public async Task GetAllNotes_ShouldMatchOnOrganisationId()
...
+
+                context.Aatfs.Add(aatf1);
+
+                await database.WeeeContext.SaveChangesAsync();
+
+                var note1 = NoteCommon.CreateNote(database, organisation1, scheme, aatf1);
+                
+                context.Notes.Add(note1);
+
+                var transferNote = NoteCommon.CreateTransferNote(database, organisation1, scheme);
```
meglenapnovoselska, line 90: you need an evidence note created, not transfer right?
is not better to set up the note with await SetupSingleNote(context, database, NoteType.EvidenceNote);

meglenapnovoselska (approved):

src/EA.Weee.DataAccess.Tests.Integration/EvidenceNote/NotesDataAccessIntegration.cs
JohnPadgettCivica: CreateNote creates a normal evidence note and CreateTransferNote creates the transfer note., they use their respective Note constructors. 

Then the assert just checks the the evidence note is retrieved.

PR #823, opened 2022-05-31
PR - 45485 - Display inline error when there are no tonnages entered

meglenapnovoselska:
A fix of the bug - In line error is not displayed when the tonnage is not entered


src/EA.Weee.Web/Areas/Scheme/Views/TransferEvidence/TransferTonnage.cshtml
```diff
@@ -55,6 +55,11 @@
         {
             var note = Model.EvidenceNotesDataList.ElementAt(i);
 
+            <div class="govuk-form-group @Html.Gds().FormGroupClass(m => m.TransferCategoryValues)">
+                <label for="TransferCategoryValues" class="govuk-visually-hidden">hidden link for navigation to tonnage error</label> <input id="CategoryValues" class="govuk-visually-hidden" />
```
JohnPadgettCivica, line 59: Think may just need to update the id from CategoryValues to TransferCategoryValues

meglenapnovoselska: it is

PR #834, opened 2022-06-06
Feature/45589 bug fix

GuidoWorkDev:
Recipient list box had a smaller font (16px) checking with UX Designers, this should be 19px.


JohnPadgettCivica:
The autocomplete sass file has been moved to content\weee\_auto-complete.scss so this change needs applying there now... maybe easiest to close this and do another PR so that we don't get merge issues

PR #847, opened 2022-06-13
45700 - PCS Add 'Outgoing transfers' tab (EXTERNAL)

meglenapnovoselska:
Added transfer out tab
Display notes from the database which are transferred
Back link works

### DIFF
--- a/src/EA.Weee.Domain/Evidence/EvidenceNoteFilter.cs
+++ b/src/EA.Weee.Domain/Evidence/EvidenceNoteFilter.cs
@@ -28,7 +28,7 @@
 
         public DateTime? EndDateSubmitted { get; set; }
 
-        public int? NoteTypeFilter { get; set; }
+        public int NoteTypeFilter { get; set; }
 
         public string FormattedSearchRef
         {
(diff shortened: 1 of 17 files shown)


src/EA.Weee.Domain/Evidence/EvidenceNoteFilter.cs
```diff
@@ -28,6 +28,8 @@ public class EvidenceNoteFilter
```
JohnPadgettCivica, line 28: Would be tempted to update the constructor of EvidenceNoteFilter to take a NoteType as we always require that.

src/EA.Weee.Domain/Evidence/EvidenceNoteFilter.cs
```diff
@@ -28,6 +28,8 @@ public class EvidenceNoteFilter
 
         public DateTime? EndDateSubmitted { get; set; }
 
+        public int? NoteTypeFilter { get; set; }
```
JohnPadgettCivica, line 31: As above I think this should be mandatory unless we use it somewhere else that doesnt require a note type

PR #853, opened 2022-06-14
Resolve issues with transfered out tab

JohnPadgettCivica:


src/EA.Weee.RequestHandlers/AatfEvidence/GetEvidenceNotesByOrganisationRequestHandler.cs
```diff
@@ -44,9 +45,20 @@ public async Task<List<EvidenceNoteData>> HandleAsync(GetEvidenceNotesByOrganisa
...
+                organisationId = request.OrganisationId;
+                schemeId = null;
+            }
+            
+            var filter = new NoteFilter(DateTime.Now.Year)
             {
-                SchemeId = scheme.Id,
+                NoteTypeFilter = new List<NoteType>() { request.NoteTypeFilter.ToDomainEnumeration<NoteType>() },
+                SchemeId = schemeId,
+                OrganisationId = organisationId,
```
meglenapnovoselska, line 61: if the request is not Transferred Out what is the organisationId be? null?

src/EA.Weee.Requests/AatfEvidence/GetEvidenceNotesByOrganisationRequest.cs
```diff
@@ -8,9 +8,13 @@
 
     public class GetEvidenceNotesByOrganisationRequest : EvidenceNoteFilterBaseRequest
     {
+        public bool TransferredOut { get; private set; }
+
         public GetEvidenceNotesByOrganisationRequest(Guid organisationId, 
             List<NoteStatus> allowedStatuses,
-            int complianceYear, NoteType noteTypeFilter)
+            int complianceYear, 
+            NoteType noteTypeFilter,
```
meglenapnovoselska, line 16: does it need to be a list? in case you want to see both types on the same tab

JohnPadgettCivica: yeah will make that update when we know how the tabs should be working for transfer notes

PR #873, opened 2022-06-17
View Draft Evidence Note Complete

NateHogan:
Due to upstream issues, previous branch was abandoned


JohnPadgettCivica:
Happy with this to be accepted, if you can make the following changes in another PR

Move the _ViewTransferNotePartial to the shared folder in the Scheme area, as it will only ever be used by a scheme
Can remove  @Html.Hidden("categoryid_to_calculate" + categoryCount, Model.TotalCategoryValues.ElementAt(categoryCount).CategoryId) as that is only used by the javascript total calculation which we don’t have on this screen

Order the notes by reference desc (or same order as the tabs)?

Unit tests....

PR #874, opened 2022-06-20
Add a boolean flag to determine whether the compliance year check is mandatory or not 

meglenapnovoselska:


src/EA.Weee.Web/Areas/Aatf/ViewModels/SubmittedDatesFilterViewModel.cs
```diff
@@ -20,7 +14,7 @@ public class SubmittedDatesFilterViewModel
 
         [Display(Name = "Submitted date")]
         [DataType(DataType.Date)]
-        [EvidenceNoteEndDate(nameof(StartDate))]
+        [EvidenceNoteEndDate(nameof(StartDate), false)]
```
JohnPadgettCivica, line 17: Think the SubmittedDatesFilterViewModel_EndDate_ShouldHaveStartDateAttribute unit test needs to be updated to check for false

meglenapnovoselska: okay

meglenapnovoselska: Very good point, I was testing the wrong model

PR #878, opened 2022-06-20
Feature/43846 task add compliance year filter external

GuidoWorkDev:
renamed all SelectedComplianceYear to ComplianceYear + unit tests


JohnPadgettCivica:
Tests are fine. Can you rename the view model compliance year property to SelectedComplianceYear. This field is named this hidden fields in the views and routes so we should keep as is.

PR #886, opened 2022-06-21
PR - Alignment of  total reuse transferred(tonnes) column

meglenapnovoselska:
bug 46302


src/EA.Weee.Web/Areas/Scheme/Views/TransferEvidence/_ViewTransferEvidenceNotePartial.cshtml
```diff
@@ -113,10 +113,10 @@
                 <th class="govuk-table__header govuk-!-text-align-left" style="width: 45%;">
                     Category
                 </th>
-                <th class="govuk-table__header govuk-!-text-align-right" style="width: 20%;">
+                <th class="govuk-table__header govuk-!-text-align-left" style="width: 18%;">
```
JohnPadgettCivica, line 116: Think we should maybe keep the column widths as they were, as think Nathan changed these previously

PR #890, opened 2022-06-22
Fix date error message

JohnPadgettCivica:
Fix to display the display message rather than model property

### DIFF
--- a/src/EA.Weee.Web/App_Start/CustomDateBinder.cs
+++ b/src/EA.Weee.Web/App_Start/CustomDateBinder.cs
@@ -1,6 +1,7 @@
 ﻿namespace EA.Weee.Web.App_Start
 {
     using System;
+    using System.Linq;
     using System.Web.Mvc;
 
     public class UKDateTimeModelBinder : IModelBinder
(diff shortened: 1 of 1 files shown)


src/EA.Weee.Web/App_Start/CustomDateBinder.cs
```diff
@@ -32,7 +33,7 @@ public object BindModel(ControllerContext controllerContext, ModelBindingContext
             }
             catch (Exception)
             {
-                bindingContext.ModelState.AddModelError(bindingContext.ModelName, string.Format("\"{0}\" is invalid.", bindingContext.ModelName));
+                bindingContext.ModelState.AddModelError(bindingContext.ModelName, $"{bindingContext.ModelMetadata.DisplayName} is invalid");
```
meglenapnovoselska, line 36: wondering whether the ModelMetadata.PropertyName to be included in the error too, up to you

JohnPadgettCivica: It turns out it displays "Start date is invalid" so looks ok

meglenapnovoselska: alright

PR #975, opened 2022-07-20
Feature/47929 bug fix accessibility element id not unique

GuidoWorkDev:
there were 2 input elements with same id="tab" on AATF manage evidence notes tabs.

### DIFF
--- a/src/EA.Weee.Web/Areas/Aatf/Views/ManageEvidenceNotes/Overview/EditDraftReturnedNotesOverview.cshtml
+++ b/src/EA.Weee.Web/Areas/Aatf/Views/ManageEvidenceNotes/Overview/EditDraftReturnedNotesOverview.cshtml
@@ -18,9 +18,8 @@
 
                 <section class="govuk-tabs__panel" id="editDraftReturnedNotes">
 
-                    @using (Html.BeginForm("Index", "ManageEvidenceNotes", FormMethod.Get))
+                    @using (Html.BeginForm("Index", "ManageEvidenceNotes", new RouteValueDictionary() { { "tab", Model.ActiveOverviewDisplayOption } }, FormMethod.Get))
                     {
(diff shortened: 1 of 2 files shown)


src/EA.Weee.Web/Areas/Aatf/Views/ManageEvidenceNotes/Overview/EditDraftReturnedNotesOverview.cshtml
```diff
@@ -20,7 +20,7 @@
 
                     @using (Html.BeginForm("Index", "ManageEvidenceNotes", FormMethod.Get))
                     {
-                        @Html.Hidden("tab", Model.ActiveOverviewDisplayOption)
```
JohnPadgettCivica, line 23: @GuidoWorkDev 
I don't think this will work. The reason for the hidden input has this name is that is gets mapped to the action parameters i.e. "tab" in the index action.

Index(Guid organisationId, Guid aatfId, 
            string tab = null,
            ManageEvidenceNoteViewModel manageEvidenceNoteViewModel = null)


I think we may have to change the 
@using (Html.BeginForm("Index", "ManageEvidenceNotes", FormMethod.Get))

to something like

@using (Html.BeginForm("Index", new RouteValueDictionary() { { "tab", Model.ActiveOverviewDisplayOption } }, "ManageEvidenceNotes", FormMethod.Get))

to remember the selected tab.

GuidoWorkDev: done

JohnPadgettCivica: ta... will merge in a bit later

PR #990, opened 2022-07-27
PR - 47957- Add compliance year filter to AA View all evidence screen 

meglenapnovoselska:


src/EA.Weee.Web.Tests.Unit/Areas/Admin/Controllers/ManageEvidenceNotesControllerTests.cs
```diff
@@ -20,6 +21,16 @@
...
     public class ManageEvidenceNotesControllerTests : ManageEvidenceNotesControllerTestsBase
     {
+        private readonly DateTime currentDate;
+
+        public ManageEvidenceNotesControllerTests()
+        {
+            currentDate = new DateTime(2019, 1, 1);
+            A.CallTo(() => WeeeClient.SendAsync(A<string>._, A<GetApiUtcDate>._)).Returns(currentDate);
+
+            A.CallTo(() => SessionService.GetTransferSessionObject<object>(ManageEvidenceController.Session, SessionKeyConstant.AASelectedComplianceYear)).Returns(null);
```
JohnPadgettCivica, line 31: I've changed the way that the compliance year is returned to the manage screens when viewing notes now so we dont need this session. I'll be removing the one from the AATF manage evidence notes index action.

meglenapnovoselska: okay, is that change in dev?

src/EA.Weee.Web/Areas/Admin/Mappings/ToViewModel/ViewAllNotesMapBase.cs
```diff
@@ -16,12 +18,25 @@ protected ViewAllNotesMapBase(IMapper mapper)
```
JohnPadgettCivica, line 18: This may be a bit of a pain but this code looks very similar to ListOfNotesViewModelBase, wonder if we can reuse that rather than the ViewAllNotesMapBase class.

src/EA.Weee.Web.Tests.Unit/Areas/Admin/Controllers/ManageEvidenceNotesControllerTests.cs
JohnPadgettCivica: Yeah now when returning from viewing of notes it should use the the note compliance year to pass that back to the manage evidence controller. 

for example
 Url.Action("Index", "ManageEvidenceNotes",
            new { @pcsId = Model.SchemeId, @tab = Model.RedirectTab, @selectedComplianceYear = Model.ComplianceYear }), "Manage evidence notes")).... so when the new view screens are created we just need to use the current note compliance year rather than trying to remember the selection... 

I'll be making the change in the AATF area soon...

meglenapnovoselska: okay, i will put it.

src/EA.Weee.Web/Areas/Admin/Mappings/ToViewModel/ViewAllNotesMapBase.cs
meglenapnovoselska: yes, I think it could be re-used.

JohnPadgettCivica:
Happy to accept this one so that Sneha can get testing baring in mind the comment made in teams and just need a couple of integration tests to fix by the looks of it

image

PR #1001, opened 2022-08-02
PR - Added an Integration test for GetComplianceYearsForNotes

meglenapnovoselska:


src/EA.Weee.Web.Tests.Unit/Areas/Admin/Controllers/ManageEvidenceNotesControllerTests.cs
```diff
@@ -78,6 +83,25 @@ public async Task IndexGet_GivenDefaultAndViewAllEvidenceNotesTab_AllEvidenceNot
                      types.SequenceEqual(g.NoteTypeFilterList)))).MustHaveHappenedOnceExactly();
         }
 
+        [Theory(Skip = "TODO: be fixed")]
```
JohnPadgettCivica, line 86: ah I think the reason that the assertion isnt working here is that it doesnt get actually called in this method, all you do is pass the func to the map object so that is all you can test for. In the view model mapper you should be able to test that the func passed to the mapper is called... that or just pass the physical objects, not sure which is easier.

meglenapnovoselska: okay, I will try it

PR #1005, opened 2022-08-03
Feature/48724 edit returned evidence note transfer

GuidoWorkDev:


src/EA.Weee.Web/Areas/Scheme/Views/TransferEvidence/_ViewTransferEvidenceNotePartial.cshtml
```diff
@@ -71,11 +72,20 @@
         <p aria-labelledby="recipient" class="govuk-body weee-word-break">@Html.Raw(Model.RecipientAddress)</p>
     </div>
 
-    @if (Model.EditMode)
+    @if (Model.EditMode && Model.HasBeenReturned)
```
JohnPadgettCivica, line 75: Should this be an Or, as user should be able to edit the transfer without it being returned... looks like it wont go in unless both are true

JohnPadgettCivica (changes requested):
Think there could possibly be an AC missing 
"Recipient must be a non editable field on page <'4.25a Edit categories with recipient change'>"

JohnPadgettCivica:
One last thing, in EditEvidenceNoteRequestHandler there is a check to ensure that the recipient hasnt changed., this section of code,

 if (!EnsureTheSchemeNotChanged(evidenceNote, message.RecipientId))
            {
                throw new InvalidOperationException($"Evidence note {evidenceNote.Id} has incorrect Recipient Id to be saved");
            }

Could we do the same in EditTransferEvidenceNoteRequestHandler. Along with unit tests...

JohnPadgettCivica (approved):

PR #1012, opened 2022-08-08
PR - 48720 - PCS - Add back buttons on edit draft and returned transfer journey (EXTERNAL)

meglenapnovoselska:


JohnPadgettCivica:
@meglenapnovoselska 
Possibly need to clear the session values when entering the edit draft transfer screen as suspect the session values will override what has been saved for the note.

Think there is the same issue we had with the create transfer screen. So Edit -> Edit categories -> change categories -> click continue -> change notes, click back (without clicking continue) -> click continue, doesnt seem to remember the selection changes before having clicked the back button

PR #1036, opened 2022-08-16
Feature/49276 bug fix browser back link issue

GuidoWorkDev:
prevent browser back button from posted edit tonnages twice

### DIFF
--- a/src/EA.Weee.Web/Areas/Scheme/Controllers/OutgoingTransfersController.cs
+++ b/src/EA.Weee.Web/Areas/Scheme/Controllers/OutgoingTransfersController.cs
@@ -51,6 +51,15 @@
 
             using (var client = apiClient())
             {
+                var noteData = await client.SendAsync(User.GetAccessToken(), new GetTransferEvidenceNoteForSchemeRequest(evidenceNoteId));
+
+                if (noteData.Status != NoteStatus.Returned && noteData.Status != NoteStatus.Draft)
+                {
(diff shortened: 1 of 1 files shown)


src/EA.Weee.Web.Tests.Unit/Areas/Scheme/Controllers/OutgoingTransfersControllerTests.cs
```diff
@@ -628,14 +665,37 @@ public async Task EditTonnagesGet_GivenMappedModel_ModelShouldBeReturned()
...
         }
 
+        [Fact]
+        public async Task EditTonnagesGet_ShouldReturnManageEvidenceView()
+        {
+            //act
+            var evidenceNoteId = TestFixture.Create<Guid>();
+            var note = TestFixture.Create<TransferEvidenceNoteData>();
+            note.Status = NoteStatus.Submitted;
+            A.CallTo(() => weeeClient.SendAsync(A<string>._,
```
JohnPadgettCivica, line 688: Can do a theory
[Theory]
[ClassData(typeof(NoteStatusCoreData))] and pass Core.AatfEvidence.NoteStatus status as test parameter.

And just do if something like in the test to ignore the two status

if (status.Equals(Core.AatfEvidence.NoteStatus.Returned) ||
                status.Equals(Core.AatfEvidence.NoteStatus.Draft))
            {
                return;
            }

PR #1098, opened 2022-09-09
Unit Tests for 50447 PCS submit transfer buttons to edit draft and returned transfer external

GuidoWorkDev:


JohnPadgettCivica (changes requested):
@GuidoWorkDev 
You are missing two main tests here...

1. You need a test that the SetNoteStatusRequest request is called with the submit status and correct evidence note id
2. That TempData[ViewDataConstant.TransferEvidenceNoteDisplayNotification] is set with the correct values based on the notes status

JohnPadgettCivica (approved):

PR #1114, opened 2022-09-15
Feature/bug fix 51150 pagination not working

GuidoWorkDev:
fixed all the places where ViewBag.Page were missing and added a few extra page parameters in controller.
Now both View and Edit links will remember which page to return to.


src/EA.Weee.Web/Areas/Scheme/Views/OutgoingTransfers/EditDraftTransfer.cshtml
```diff
@@ -7,15 +7,18 @@
 
 @section full_main_content
 {
+    <p>Guido</p>
```
JohnPadgettCivica, line 10: @GuidoWorkDev  Remove this

PR #1122, opened 2022-09-16
Feature/51082 unit and integration tests

GuidoWorkDev:
implemented the CSV report filename to include approval number + related unit/integration Tests


JohnPadgettCivica (commented):
@GuidoWorkDev 
Happy to accept this but... we still need the unit tests for 

- DownloadEvidenceNotesReport (look in EvidenceReportsControllerTests for very similar tests)
- Also rather than just updating HandleAsync_GivenCsvData_CsvFileDataShouldBeReturned a separate test should be created for the AATF case. As not its not testing the non aatf case.

PR #1137, opened 2022-09-22
AATF - Delete validation

NateHogan:


src/EA.Weee.RequestHandlers/Admin/Aatf/GetAatfDeletionStatus.cs
```diff
@@ -40,6 +40,13 @@ public async Task<CanAatfBeDeletedFlags> Validate(Guid aatfId)
                 return result;
             }
 
+            var hasNotes = await aatfDataAccess.HasEvidenceNotes(aatfId);
+            if (hasNotes)
+            {
+                result |= CanAatfBeDeletedFlags.HasNotes;
```
JohnPadgettCivica, line 46: @NateHogan  To simplify things even further as we don't need to explicitly check HasNotes. Just set the existing HasData flag.

PR #1163, opened 2022-10-03
51587 - Add Filters on AA screen - unit and integration tests

meglenapnovoselska:
…er changes


src/EA.Weee.RequestHandlers/Admin/GetAllAatfsForComplianceYearHandler.cs
```diff
@@ -35,7 +36,9 @@ public async Task<List<OrganisationSchemeData>> HandleAsync(GetAllAatfsForCompli
                 throw new ArgumentException($"Aatfs not found for a compliance year {message.ComplianceYear}");
             }
 
-            return mapper.Map<List<Domain.AatfReturn.Aatf>, List<OrganisationSchemeData>>(listAatfs);
+            var result = mapper.Map<List<Domain.AatfReturn.Aatf>, List<OrganisationSchemeData>>(listAatfs);
```
JohnPadgettCivica, line 39: Wonder if we are using OrganisationSchemeData for a more generic Id, DisplayName pair, perhaps we can rename it to something more generic... Maybe EntityIdDisplayNameData or something :)

meglenapnovoselska: Yes, I agree with that. Also needs to be put in a Shared place. I will do it.

JohnPadgettCivica:
Looks good if as you say could just rename that class...

PR #1186, opened 2022-10-11
PR - 52654 - NEW - AA - Add ref ID filter to view all evidence notes tab

meglenapnovoselska:


src/EA.Weee.Integration.Tests/Builders/EvidenceNoteDbSetup.cs
```diff
@@ -60,6 +60,12 @@ public EvidenceNoteDbSetup WithAatf(Guid aatfId)
             return this;
         }
 
+        public EvidenceNoteDbSetup WithReference(int reference)
+        {
+            ObjectInstantiator<Note>.SetProperty(o => o.Reference, reference, instance);
```
JohnPadgettCivica, line 65: @meglenapnovoselska The reference is an identity column in the database so will be readonly. Suspect this is why the test isnt working. Your integration test creates the notes using the EvidenceNoteDbSetup class. When .Create() is called it will be assigned a Reference. So for example you will be able to do evidenceWithFilteredReference.Reference which will give you the Reference that was generated by SQL Server, which you can then pass to the request.

Would probably just remove the WithReference method.

meglenapnovoselska: thank you.

PR #1189, opened 2022-10-12
Delete null/zero tonnage - Integration Tests

NateHogan:


src/EA.Weee.Integration.Tests/Handlers/CreateTransferEvidenceNoteHandlerIntegrationTests.cs
```diff
@@ -273,6 +273,74 @@ public class WhenICreateADraftTransferEvidenceNoteWhereThereIsNoAvailableTonnage
             private readonly It shouldHaveCaughtInvalidOperationException = ShouldThrowException<InvalidOperationException>;
         }
 
+        [Component]
```
JohnPadgettCivica, line 276: @NateHogan Like this one... could you update the existing Create transfer note integration tests to check that when null and zero tonnage is present it isnt removed.

For example update WhenICreateADraftTransferEvidenceNote.

PR #1202, opened 2022-10-14
feature/52843 PBS/PCS transfer notes report EXTERNAL

GuidoWorkDev:


src/EA.Weee.Web/Areas/Scheme/Controllers/ManageEvidenceNotesController.cs
```diff
@@ -327,6 +327,21 @@ public async Task<ActionResult> DownloadEvidenceNotesReport(Guid pcsId, int comp
             }
         }
 
+        [HttpGet]
+        public async Task<ActionResult> DownloadEvidenceTransfersReport(Guid pcsId, int complianceYear)
+        {
+            using (var client = apiClient())
```
JohnPadgettCivica, line 333: @GuidoWorkDev unit tests :)

src/EA.Weee.DataAccess/StoredProcedure/TransferNoteData.cs
```diff
@@ -23,7 +23,7 @@ public class TransferNoteData
 
         public string EvidenceNoteReference { get; set; }
 
-        public DateTime EvidenceNoteApprovalDate { get; set; }
+        public DateTime? EvidenceNoteApprovalDate { get; set; }  // sql query moanied about this not being nullable
```
JohnPadgettCivica, line 26: @GuidoWorkDev  I cant see why this would be the case may have to look at what data causes this as all evidence notes for transfer notes should have an approval date. Perhaps funny data. In any case this should really be as it was.

Can you make sure you are complete up to date with the database updates.

JohnPadgettCivica: @GuidoWorkDev will need to find out why this was needed... Will pick up on Tuesday

JohnPadgettCivica (changes requested):
Need to check the data issue...

src/EA.Weee.DataAccess/StoredProcedure/TransferNoteData.cs
GuidoWorkDev: In the case on my dataset there were some nulls in the EvidenceNoteApprovalDate column which raised errors in the handler.  it can fix my dataset (remove the NULL values) and I will rollback to original type.

src/EA.Weee.Web/Areas/Scheme/Controllers/ManageEvidenceNotesController.cs
GuidoWorkDev: I have added unit tests to cover this code

JohnPadgettCivica (approved):

PR #1214, opened 2022-10-19
PR - 53366 - Message displayed when the search element did not match any results is different

meglenapnovoselska:


JohnPadgettCivica:
@meglenapnovoselska I think this new message should only be displayed when a search has been performed... I thnk the AATF screen may make a check and display different messages

meglenapnovoselska:
> 



> @meglenapnovoselska I think this new message should only be displayed when a search has been performed... I thnk the AATF screen may make a check and display different messages

I thought so the same, but I've checked the ViewAllOtherEvidenceOverview view and in case Model.EvidenceNotesDataList.Any() == false, it gets set to _NoResultsPartial.cshtml with the message  >There are no search results with those search terms. Please try again with another search term. Same is valid for the AA Internal -> Evidence Note page. The "old" messages seems to be dropped. 
image

JohnPadgettCivica:
@meglenapnovoselska That looks like a bug to me... That should be "<p>There are currently no evidence notes </p>" for that view if you could change it back?

It looks like in _NoResultsPartial should render the <p>There are no search results with those search terms. Please try again with another search term.</p> message when a search has been performed. Not sure why that isnt working. Maybe not detecting a search has been performed...

meglenapnovoselska:
> 

okay, I will push the fixes for 
- AATF screen
- AA Evidence screen
- AA Transfer screen

can't think of any other places the search filter being used

PR #1250, opened 2022-11-01
Fix no results format issue

JohnPadgettCivica:


meglenapnovoselska:
just thinking that text   <p> Showing 0 - 0 of 0 </p> is in other PSC and Aatf .cshtml files too, with 'no results' message with colspan=6
whether they need to be updated or not?

JohnPadgettCivica:
> just thinking that text
> 
> Showing 0 - 0 of 0
> 
> is in other PSC and Aatf .cshtml files too, with 'no results' message with colspan=6
> whether they need to be updated or not?

Good point will have a look...

JohnPadgettCivica:
> just thinking that text
> 
> Showing 0 - 0 of 0
> 
> is in other PSC and Aatf .cshtml files too, with 'no results' message with colspan=6
> whether they need to be updated or not?

Updated...

meglenapnovoselska (approved):

PR #1251, opened 2022-11-01
Adds filters to AA - View evidence transfers tab

NateHogan:


src/EA.Weee.Web/Areas/Admin/Controllers/ManageEvidenceNotesController.cs
```diff
@@ -116,9 +116,11 @@ public async Task<ActionResult> ViewEvidenceNoteTransfer(Guid evidenceNoteId, in
                    result, TempData[ViewDataConstant.TransferEvidenceNoteDisplayNotification], this.User)
                 {
                     OpenedInNewTab = openedInNewTab,
-                    Page = page
+                    Page = page,
```
JohnPadgettCivica, line 119: @NateHogan I think this model and transfer object has a query string property can we move to use the model mapper rather than the view bag...

PR #1257, opened 2022-11-02
PR - 53666 - AATF - Remembering Search Reference When user goes back  

meglenapnovoselska:


JohnPadgettCivica:
@meglenapnovoselska Looks like there is a compilation warning on this one...

meglenapnovoselska:
> @meglenapnovoselska Looks like there is a compilation warning on this one...

okay, I will check

meglenapnovoselska:
> @meglenapnovoselska Looks like there is a compilation warning on this one...

the QueryString property was added twice to the same class.

PR #1296, opened 2023-03-08
Change to clean the query string that causes issues in production env…

JohnPadgettCivica:
…ironment


sreedhar-civica:
This changes already merged into the Release Candidate 3.10

PR #1343, opened 2023-04-20
Content changes implemented for the return aatf.

sreedhar-civica:


src/EA.Weee.Web/Areas/AatfReturn/Views/SelectReportOptionsDeselect/Index.cshtml
```diff
@@ -16,27 +16,30 @@
...
-            {
-                @deselectedOption.AlternativeDescription
-            }
-            else
+    <div class="govuk-body">
+        <ul>
+            @foreach (var deselectedOption in Model.ReportOnQuestions.Where(r => r.DeSelected))
             {
-                @deselectedOption.Title
+                <li style="padding-top: 10px; list-style-type: none;">
```
JohnPadgettCivica, line 30: I think you could use ol here... as I think that has list-style-type: none defined. Also for the padding you could use govuk-!-padding-top-3 as that has defined padding.

PR #1354, opened 2023-05-04
Change to fix validation where available re-used is zero as it should…

JohnPadgettCivica:
…nt be validated


sreedhar-civica:
Created new pull request using release/3_12

PR #1355, opened 2023-05-04
Prototype filter changes for issue with production filters

JohnPadgettCivica:
This is just an idea of how we could reduce the query string size. The same process would need to be completed for the other filters and perhaps the external screens (may be less of an issue there as they have less filters)


sreedhar-civica:
Closed this pull request. The changes implemented on another branch.

PR #1358, opened 2023-05-12
Access Denied error when attempting to ‘View’ an evidence note issue …

sreedhar-civica:


JonnyRush (approved):
Looks good, haven't checked locally but share screened with Sreedhar testing it.

PR #1359, opened 2023-05-30
I975212 unable to submit quarter data

sreedhar-civica:


stephen-gowen-civica:
After explanation by Sreedhar, I approve this change

PR #1409, opened 2024-08-22
70856 organisation transaction service

JohnPadgettCivica:
This adds some database and EF config for https://dev.azure.com/civica-digital/EA%20-%20WEEE/_workitems/edit/70856.

It also adds some structure to enable us to store partially completed organisation data.


src/EA.Weee.Core/Organisations/OrganisationTransactionData.cs
```diff
@@ -0,0 +1,14 @@
...
+    using System;
+
+    public class OrganisationTransactionData
+    {
+        public Guid? Id {get; set; }
+
+        public bool? PreviousNpwdAccount { get; set; }
+
+        // each screen view model
+        public OrganisationDetails OrganisationDetails { get; set; }
```
JohnPadgettCivica, line 12: This is just an example.

PR #1414, opened 2024-08-27
Schema updates use notification address

JohnPadgettCivica:


src/EA.Weee.DataAccess/Mappings/DirectRegistrantMapping.cs
```diff
@@ -0,0 +1,22 @@
...
+        public DirectRegistrantConfiguration()
+        {
+            ToTable("DirectRegistrant", "Producer");
+
+            HasKey(e => e.Id);
+
+            Property(e => e.RowVersion).IsRowVersion();
+
+            HasRequired(e => e.Organisation).WithMany().HasForeignKey(e => e.OrganisationId);
+            HasOptional(e => e.SICCode).WithMany().HasForeignKey(e => e.SICCodeId);
```
will-hope98, line 17: I may not be remembering this correctly, but did we not say during estimation we weren't including the SIC code?

src/EA.Weee.Domain/Organisation/Address.cs
```diff
@@ -134,6 +140,19 @@ private set
...
         }
 
+        public string WebAddress
+        {
+            get => webAddress;
+            private set
+            {
+                if (value != null && value.Length > 50)
+                {
+                    throw new InvalidOperationException(string.Format(("Web address cannot be greater than 50 characters")));
```
will-hope98, line 150: Maybe slightly picky but should it not be "Website address"

src/EA.Weee.DataAccess/Mappings/DirectRegistrantMapping.cs
JohnPadgettCivica: It could be making a return :)

src/EA.Weee.Domain/Organisation/Address.cs
JohnPadgettCivica: @will-hope98  fixed.

PR #1433, opened 2024-09-03
71097 - TO1.0 "Partnership" Screen

seanciv:


src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -438,6 +439,44 @@ public async Task<ActionResult> PartnershipDetails(PartnershipDetailsViewModel m
...
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> PartnerDetails(PartnerViewModel model, string action)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            if (model.PartnerModels.Count > maxPartnersAllowed)
+            {
```
JohnPadgettCivica, line 452: @seanciv I cleaner way of doing this is for PartnerViewModel to implement IValidatableObject, which will get called during model binding, but what you have done works...

JohnPadgettCivica:
@seanciv As part of the completion of the journey the data captured by this screen needs to be persisted to the proper database tables. This happens in CompleteOrganisationTransactionHandler so we need to take the captured data and populate the AdditionalCompanyDetails tables, similar to how the other entities are created.

There are unit and integration tests in these classes that need to be updated.
CompleteOrganisationTransactionHandlerTests
CompleteOrganisationTransactionHandlerIntegrationTests

JohnPadgettCivica:
@seanciv Some conflicts here that will probably be worth merging in as they affect the two test classes you are updating.

src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -624,6 +625,44 @@ private object CastToSpecificViewModel(ExternalOrganisationType? organisationTyp
             }
         }
 
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<ActionResult> PartnerDetails(PartnerViewModel model, string action)
+        {
+            if (!ModelState.IsValid)
```
JohnPadgettCivica, line 632: @seanciv Noticed this AC
"Then the following fields are mandatory: 'First name', 'Last name' for the first 2 partners (for partner 3 onwards, the fields are optional)" is this covered? Might need an alternative PartnerModel model that doesnt have those fields as required. Could either add an additional property to PartnerViewModel to store the non mandatory field ones or define some interface between the two alternative PartnerModel types...  thats only if this currently doesnt satisfy the AC or you can think of an alternative.

src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
seanciv: Thanks, even if its cleaner a few other hurdles came up that I dealt with before this. If its "fine" as is we can merge, if not, I can spend some time tomorrow on it

PR #1440, opened 2024-09-09
76245 company registration number api

will-hope98:


JohnPadgettCivica:
Update the below so that this value is passed from config rather than hardcoded just in case it changes.
image
The configuration service is passed to the controller so you can add CompaniesHouseReferencePath  to  IAppConfiguration and AppConfiguration and you will be able to access it though the configurationService.

PR #1442, opened 2024-09-09
70929 task list

seanciv:


src/EA.Weee.Web/Areas/Producer/Controllers/ProducerController.cs
```diff
@@ -18,9 +33,58 @@ public ActionResult Index()
 
         [SmallProducerSubmissionContext]
         [HttpGet]
-        public ActionResult TaskList()
+        public async Task<ActionResult> TaskList()
         {
-            return View();
+            var submission = SmallProducerSubmissionData.CurrentSubmission;
+
+            var model = new TaskListViewModel()
```
JohnPadgettCivica, line 40: @seanciv This looks good in the new stories if we are to try and use the IMap, so in theory we pass data to class and have a mapper to generate a viewmodel making it a bit easier to test.

PR #1443, opened 2024-09-09
71096 sole trader

stevelee-civica:
71096 add sole trader screen
add unit tests
update navigations


src/EA.Weee.Core/Organisations/SoleTraderViewModel.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace EA.Weee.Core.Organisations
+{
+    using System.ComponentModel;
+    using System.ComponentModel.DataAnnotations;
+    using EA.Weee.Core.DataStandards;
+    public class SoleTraderViewModel
+    {
+        [Required]
+        [DisplayName("First name")]
+        [StringLength(CommonMaxFieldLengths.DefaultString)]
```
JohnPadgettCivica, line 10: @stevelee-civica have a look at the latest partner view model as I think they use different length fields, think there is a Firstname and lastname one

JohnPadgettCivica:
@stevelee-civica the organisation transaction service will need to be updated to persist this data I believe. 

Also as we now have this new screen in between the Type screen and Organisation details screen, clicking the Back link from organisation details will need to take us to the Partner details screen.

This line 
image
Will have to be conditional based on the Model.OrganisationType value. It also looks broken for Partner.

src/EA.Weee.Web.Tests.Unit/Controllers/OrganisationRegistrationControllerTests.cs
```diff
@@ -1435,5 +1435,42 @@ public async Task PartnerDetails_Post_AddErrorIfMoreThanTenPartners()
 
             var model = result.Model as PartnerViewModel;
         }
+
+        [Fact]
+        public async Task SoleTraderDetails_Get_ReturnsView()
+        {
+            // Act
+            var result = await controller.SoleTraderDetails() as ViewResult;
```
JohnPadgettCivica, line 1443: @stevelee-civica There should also be a test to check that when the 

var existingTransaction = await transactionService.GetOrganisationTransactionData(User.GetAccessToken());

Call is made in the GET method that that existing model is returned to the view.

stevelee-civica: added unit test

src/EA.Weee.Core/Organisations/SoleTraderViewModel.cs
stevelee-civica: updated

src/EA.Weee.Web.Tests.Unit/Controllers/OrganisationRegistrationControllerTests.cs
stevelee-civica: updated organisation transaction service to handle SoleTraderViewModel

PR #1444, opened 2024-09-09
76972 - The back button on the "Organisation details" page goes strai…

seanciv:
…ght back to "Organisation type"

### DIFF
--- a/src/EA.Weee.Web/Views/OrganisationRegistration/OrganisationDetails.cshtml
+++ b/src/EA.Weee.Web/Views/OrganisationRegistration/OrganisationDetails.cshtml
@@ -9,7 +9,7 @@
 }
 @section full_main_content
 {
-    @(this.WeeeGds().BackLink(@Url.UrlFor<OrganisationRegistrationController>(a => a.PartnerDetails())))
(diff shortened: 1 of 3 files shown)


src/EA.Weee.Web/Views/OrganisationRegistration/OrganisationDetails.cshtml
```diff
@@ -9,7 +9,7 @@
 }
 @section full_main_content
 {
-    @(this.WeeeGds().BackLink(@Url.UrlFor<OrganisationRegistrationController>(a => a.Type())))
+    @(this.WeeeGds().BackLink(@Url.UrlFor<OrganisationRegistrationController>(a => a.PartnerDetails())))
```
JohnPadgettCivica, line 12: @seanciv I think this one needs to go back to the screen relevant for the type of organisation, partner goes to partner details, registered company goes to type and sole trader goes to the other (but @stevelee-civica  is working on that)

Think there should be a Model.Organisation type so we probably have to do something like

    @if (Model.OrganisationType == ExternalOrganisationType.Partnership)
    {
// go to partner
    } else if (Model.OrganisationType == ExternalOrganisationType.RegisteredCompany)
{
// go to type
}

or have a new redirect action and do similar in that, which would probably be better as it can be tested.

PR #1445, opened 2024-09-09
Add in example producer edit details process

JohnPadgettCivica:


willdot:
> @seanciv @willdot @stevelee-civica The changes in here are the kind of process I see us following for the upcoming J3 screens. Have a look and let me know if you have questions. Note that the Edit request handler isn't in place and there are more integration tests and unit tests but this is the type of flow I think we can use.

Hey John 👋 

Not for me I guess 🤣

PR #1470, opened 2024-09-18
71111 org tabs

seanciv:


src/EA.Weee.Web/Areas/Producer/Views/Producer/ViewOrganisation/ContactDetails.cshtml
```diff
@@ -0,0 +1,39 @@
...
+@model OrganisationDetailsTabsViewModel
+
+@{
+    var @title = "Contact details";
+    ViewBag.Title = @title + " - GOV.UK";
+}
+@section full_main_content{
+    <div class="govuk-!-padding-bottom-8"></div>
+    <div class="govuk-tabs">
+        @Html.Partial("ViewOrganisation/_OrganisationDetailsTabs", Model)
```
JohnPadgettCivica, line 15: @seanciv  Sorry should have mentioned and didnt notice that the scheme used javascript. Annoyingly as this is an external screen it will also need to work without javascript so probably best we dont use the javascript method on here.  This one C:\repos\prsd-weee\src\EA.Weee.Web\Areas\Aatf\Controllers\ManageEvidenceNotesController.cs doesnt use javascript I believe it just uses the full views rather than partials.

src/EA.Weee.Web/Areas/Producer/Mappings/ToViewModel/OrganisationViewModelMap.cs
```diff
@@ -0,0 +1,47 @@
...
+    using EA.Weee.Core.Organisations.Base;
+
+    public class OrganisationViewModelMap : IMap<SmallProducerSubmissionData, OrganisationViewModel>
+    {
+        public OrganisationViewModel Map(SmallProducerSubmissionData source)
+        {
+            return new OrganisationViewModel()
+            {
+                Address = new Core.Organisations.ExternalAddressData
+                {
```
JohnPadgettCivica, line 15: @seanciv There is actually a mapper for this ReverseAddressMap. So you can inject the mapper into the class and use it similar to EditOrganisationDetailsMap.

src/EA.Weee.Web/Areas/Producer/Views/Producer/ViewOrganisation/ContactDetails.cshtml
```diff
@@ -0,0 +1,31 @@
...
+@using MvcNavigationHelpers
+
+@using EA.Weee.Web.Areas.Producer.ViewModels
+
+@model OrganisationDetailsTabsViewModel
+
+@{
+    var @title = "Contact details";
+    ViewBag.Title = @title + " - GOV.UK";
+}
```
JohnPadgettCivica, line 11: @seanciv  think we need a back button

JohnPadgettCivica:
@seanciv  looks good if Wave is happy :)

PR #1491, opened 2024-09-26
71102 org found screen

will-hope98:


src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -408,9 +408,11 @@ public async Task<ActionResult> OrganisationDetails(OrganisationViewModel model)
 
             var existingOrgs = await GetExistingOrganisations(model);
 
-            if (existingOrgs.Any())
+            if (existingOrgs.Organisations.Any())
             {
-                return await OrganisationFound(existingOrgs);
+                TempData["FoundOrganisations"] = existingOrgs;
```
JohnPadgettCivica, line 413: @will-hope98  I think we can test that TempData has been set also, I think by controller.TempData["FoundOrganisations"].should().be.....

PR #1492, opened 2024-09-26
change content

stevelee-civica:
add conditional render

### DIFF
--- a/src/EA.Weee.Web.Tests.Unit/Areas/Producer/Controllers/ProducerSubmissionControllerTests.cs
+++ b/src/EA.Weee.Web.Tests.Unit/Areas/Producer/Controllers/ProducerSubmissionControllerTests.cs
@@ -261,7 +261,7 @@
 
             // Assert
             result.Should().NotBeNull();
-            result.RouteValues["action"].Should().Be("CheckAnswers");
+            result.RouteValues["action"].Should().Be("TaskList");
             result.RouteValues["controller"].Should().Be("Producer");
             A.CallTo(() => weeeClient.SendAsync(A<string>._, request)).MustHaveHappenedOnceExactly();
         }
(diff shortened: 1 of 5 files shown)


src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -689,6 +689,8 @@ public async Task<ActionResult> ContactDetails()
             var countries = await GetCountries();
             model.AddressData.Countries = countries;
 
+            model.HasAuthorisedRepresentitive = existingTransaction?.AuthorisedRepresentative == YesNoType.Yes;
```
JohnPadgettCivica, line 692: @stevelee-civica  can you create a unit tests to ensure this value is set correctly...

src/EA.Weee.Web.Tests.Unit/Areas/Producer/Controllers/ProducerSubmissionControllerTests.cs
```diff
@@ -261,7 +261,7 @@ public async Task EditContactDetails_Post_ValidModel_ShouldRedirectToTaskList()
 
             // Assert
             result.Should().NotBeNull();
-            result.RouteValues["action"].Should().Be("TaskList");
+            result.RouteValues["action"].Should().Be("CheckAnswers");
```
will-hope98, line 264: @stevelee-civica does this test pass now?  I'm not sure this change is needed

stevelee-civica: restored the changes
it fails the test case now

src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
stevelee-civica: updated and added assertions on without existing transaction

PR #1501, opened 2024-10-01
Update to allow removal of partner and fix hidden text on an old screen

JohnPadgettCivica:


seanciv:
@JohnPadgettCivica All looks good to me!

PR #1504, opened 2024-10-01
Refactor of the mapping for submission data

JohnPadgettCivica:


src/EA.Weee.RequestHandlers/Mappings/SmallProducerSubmissionHistoryDataMap.cs
```diff
@@ -0,0 +1,91 @@
...
+        public SmallProducerSubmissionHistoryDataMap(IMapper mapper)
+        {
+            this.mapper = mapper;
+        }
+
+        public SmallProducerSubmissionHistoryData Map(DirectProducerSubmissionSource source)
+        {
+            Condition.Requires(source).IsNotNull();
+
+            return new SmallProducerSubmissionHistoryData()
```
seanciv, line 27: @JohnPadgettCivica so many short hand conditions here makes it difficult to read. It's minor, but is there not a way to check once if CurrentSubmission is null/ empty (by allowing it to be nullable further up) then build for DirectRegistrant.Organisation else build for DirectProducerSubmission.CurrentSubmission

JohnPadgettCivica: @seanciv Updated to make more readable

seanciv:
Look good to me! @JohnPadgettCivica

PR #1515, opened 2024-10-08
76765 download pdf

will-hope98:


src/EA.Weee.Web/Views/Shared/_DownloadSubmissionPdfLayout.cshtml
```diff
@@ -0,0 +1,266 @@
...
+
+        .govuk-full-printable-width {
+            width: 100% !important;
+            display: flex;
+        }
+
+            .govuk-full-printable-width span:nth-child(1) {
+                width: 100%;
+            }
+
```
JohnPadgettCivica, line 17: @will-hope98  Bit of a pain but would you be able to see if you can move the head and footer into partials that can be re-used across this and the evidence note PDF?

will-hope98: Moved to use one partial

PR #1522, opened 2024-10-09
76931 - TO3.0 - Continue submission (submission completed)

seanciv:


src/EA.Weee.Web/Areas/Producer/Controllers/ProducerController.cs
```diff
@@ -54,12 +54,24 @@ public ActionResult Index()
...
+            return View(SmallProducerSubmissionData.OrganisationData.Id);
+        }
+
         [SmallProducerSubmissionContext]
         [HttpGet]
         public async Task<ActionResult> TaskList()
         {
             var submission = SmallProducerSubmissionData.CurrentSubmission;
 
+            if (submission.HasPaid && submission.Status == SubmissionStatus.Submitted)
```
JohnPadgettCivica, line 70: @seanciv  The issue with this is that the user will be able to directly access the pages if they have the URL favourited for example, or just by previous history access. I added some notes into the story about creating an attribute that we can associate with all the actions which would prevent the user being able to "back door" into the screen.

src/EA.Weee.Web/Areas/Producer/Views/Producer/AlreadySubmittedAndPaid.cshtml
```diff
@@ -0,0 +1,26 @@
...
+{
+    @(this.WeeeGds().BackLink(@Url.UrlFor<EA.Weee.Web.Areas.Scheme.Controllers.HomeController>(a => a.ChooseActivity(Model))))
+
+    <header class="hgroup">
+        <h1 class="govuk-heading-l">@title</h1>
+    </header>
+
+    <p class="govuk-body">
+        You have already submitted your producer registration for the current year.
+        If you would like to view you current year submission,
```
JohnPadgettCivica, line 19: @seanciv  Thinking back to some screen reader issues we have had might be worth removing any newlines etc in the cshml and the text can just flow in a normal P.

JohnPadgettCivica:
@seanciv  looks good :) Just minor comment on the HTML as think screenreader will have a moan about newlines

PR #1527, opened 2024-10-10
add Selling technique type string

stevelee-civica:
add Selling technique type string to ui and pdf

### DIFF
--- a/src/EA.Weee.Web/Areas/Producer/ViewModels/SellingTechniqueViewModel.cs
+++ b/src/EA.Weee.Web/Areas/Producer/ViewModels/SellingTechniqueViewModel.cs
@@ -30,21 +30,6 @@
             throw new InvalidOperationException();
         }
 
-        public string ToSellingTechniqueTypeString() 
-        {
-            switch (ToSellingTechniqueType())
-            {
-                case SellingTechniqueType.DirectSellingToEndUser:
-                    return "Direct selling to end user (mail, order, internet etc)";
-                case SellingTechniqueType.IndirectSellingToEndUser:
-                    return "Indirect selling (other)";
-                case SellingTechniqueType.Both:
-                    return "Direct selling to end user (mail, order, internet etc) and Indirect selling (other)";
-                default:
-                    return string.Empty;
-            }
-        }
-
         public static SellingTechniqueViewModel FromSellingTechniqueType(SellingTechniqueType? type)
         {
             return new SellingTechniqueViewModel
--- a/src/EA.Weee.Web/Areas/Producer/ViewModels/SellingTechniqueViewModel.cs
+++ b/src/EA.Weee.Web/Areas/Producer/ViewModels/SellingTechniqueViewModel.cs
@@ -30,6 +30,21 @@
             throw new InvalidOperationException();
         }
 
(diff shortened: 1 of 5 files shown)


src/EA.Weee.Web/Areas/Producer/ViewModels/SellingTechniqueViewModel.cs
```diff
@@ -30,6 +30,21 @@ public SellingTechniqueType ToSellingTechniqueType()
             throw new InvalidOperationException();
         }
 
+        public string ToSellingTechniqueTypeString()
```
seanciv, line 33: Can we not use Display() attribute on enum and use the GetDisplayName helper in cshtml?

JohnPadgettCivica: @stevelee-civica  @seanciv  It would be worth clarifying with James that if both is selected we need to display both options or "Both". When selecting it on the EE screen Both isn't an option that they don't select both its inferred by the fact you select both radio buttons.

stevelee-civica: Confirmed with James
possible display texts
1: "Direct selling to end user (mail, order, internet etc)"
2: "Indirect selling (other)"
3: "Direct selling to end user (mail, order, internet etc) and Indirect selling (other)"

JohnPadgettCivica: @stevelee-civica  Ok as Sean mentions rather than hard coding the text would it be possible to use the display helpers to create that text. And need unit test for this method...

stevelee-civica: refactored to using Html.DisplayNameFor

src/EA.Weee.Web/Areas/Producer/Views/Shared/_CheckAnswersContentPartial.cshtml
```diff
@@ -157,6 +160,26 @@
...
         }
+        <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header govuk-!-text-align-left" style="width: 28%;">Selling technique(s) that you use</th>
+            <td class="govuk-table__cell" colspan="2">
+                @switch(Model.EeeData.SellingTechnique.ToSellingTechniqueType())
+                {
+                    case SellingTechniqueType.DirectSellingToEndUser:
+                        sellingTechniqueType = Html.DisplayNameFor(m => m.EeeData.SellingTechnique.IsDirectSelling).ToString();
+                        break;
+                    case SellingTechniqueType.IndirectSellingToEndUser:
```
JohnPadgettCivica, line 171: @stevelee-civica This should be done in code. There is EnumHelper that I think should do what we need, does that not work? The model method can be unit tested, this cant...

stevelee-civica: updated and added unit tests

PR #1532, opened 2024-10-11
Api client updates

JohnPadgettCivica:


src/EA.Weee.Api.Client/CompaniesHouseClient.cs
```diff
@@ -24,40 +24,45 @@ public CompaniesHouseClient(
...
 
-            this.httpClient = httpClientFactory.CreateHttpClientWithCertificate(baseUrl, config, logger, certificate);
-            
+            this.httpClient = httpClientFactory.CreateHttpClient(baseUrl, config, logger);
             this.retryPolicy = retryPolicy;
             this.jsonSerializer = jsonSerializer;
+            this.tokenProvider = tokenProvider;
         }
 
         public async Task<DefraCompaniesHouseApiModel> GetCompanyDetailsAsync(string endpoint, string companyReference)
```
seanciv, line 44: Just some comments about errors, If I was calling this method and it returns null it wouldn't be obvious to me why, example at "if (!IsValidCompanyReference(companyReference))" it returns null but the reason is known. But maybe NULL is more useful to return, not sure. 

Another thing is you throw BadRequestException("Companies house bad request"); then immediately catch it in the same method, don't log it, and return null. If the intention is to return null, there is no need to throw, then catch, then return null but to just return null in the "response.StatusCode" check. I am also not sure how useful null will be to the caller if the status is NOTFOUND, and if null is not returned and the exception bubbles up, "Bad Request" just hides "not found" as an error. Which is confusing if it isn't logged

stevelee-civica (approved):

src/EA.Weee.Api.Client/CompaniesHouseClient.cs
JohnPadgettCivica: @stevelee-civica  are you happy with my updates?

PR #1548, opened 2024-10-17
70968 internalsubmissions

seanciv:


src/EA.Weee.Web/Areas/Admin/Controllers/ProducerSubmissionController.cs
```diff
@@ -1,13 +1,143 @@
...
+        }
+
+        [HttpGet]
+        public async Task<ActionResult> AddPaymentDetails()
         {
             return View();
         }
+
+        [HttpGet]
+        public async Task<ActionResult> RemoveSubmission()
```
JohnPadgettCivica, line 133: @seanciv Wont need to worry about this one right now but for info the Remove will probably need to hook into the Admin.ProducersController.ConfirmRemoval action. Which means we probably need to return the registeredProducerId from the handler...

src/EA.Weee.Web/Areas/Admin/Views/ProducerSubmission/RemoveSubmission.cshtml
```diff
@@ -0,0 +1,13 @@
+﻿@{
+    var title = "Add payment details";
```
will-hope98, line 2: Not important, but might be worth changing the title here in case testers complain

will-hope98 (approved):

src/EA.Weee.Web/Areas/Admin/Controllers/ProducerSubmissionController.cs
seanciv: @JohnPadgettCivica I am going to leave this here because AC states: 
"When I click the 'Remove [YEAR] submission' link
Then I am taken to the Remove submission screen (#77590)"
Which first takes the user to a new screen

PR #1585, opened 2024-10-31
fix CompanyRegistrationNumber not saved to DirectProducerSubmissionHistory

stevelee-civica:
https://dev.azure.com/civica-digital/EA%20-%20WEEE/_workitems/edit/78573/

CompanyRegistrationNumber not saved to DirectProducerSubmissionHistory
so when finishing the submission it will use null to set to organisation table

### DIFF
--- a/src/EA.Weee.Integration.Tests/Handlers/DirectRegistrant/EditOrganisationDetailsRequestHandlerIntegrationTests.cs
+++ b/src/EA.Weee.Integration.Tests/Handlers/DirectRegistrant/EditOrganisationDetailsRequestHandlerIntegrationTests.cs
@@ -157,7 +157,7 @@
                 var address = fixture.Build<AddressData>().With(a => a.CountryId, country.Id).Create();
 
                 request = new EditOrganisationDetailsRequest(directRegistrant.Id, "New company name",
-                    "New trading name", address, "New brand names");
+                    "New trading name", address, "New brand names", "Reg1234");
 
                 OrganisationUserDbSetup.Init().WithUserIdAndOrganisationId(UserId, directRegistrant.OrganisationId).Create();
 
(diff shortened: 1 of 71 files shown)


src/EA.Weee.Integration.Tests/Handlers/DirectRegistrant/EditOrganisationDetailsRequestHandlerIntegrationTests.cs
```diff
@@ -157,7 +157,7 @@ public static IntegrationTestSetupBuilder LocalSetup()
                 var address = fixture.Build<AddressData>().With(a => a.CountryId, country.Id).Create();
 
                 request = new EditOrganisationDetailsRequest(directRegistrant.Id, "New company name",
-                    "New trading name", address, "New brand names");
+                    "New trading name", address, "New brand names", "Reg1234");
```
JohnPadgettCivica, line 160: @stevelee-civica  we shouldn't be updating this request to take the company number as its a non editable field. 

Whats the underlaying cause of the bug is it being overwritten somewhere? I would suggest we create a script to remove the CompanyRegistrationNumber column from the DirectProducerSubmissionHistory and remove it from the Domain class as it shouldnt be edited from the original organisation value.

stevelee-civica: AddSignatoryAndCompleteRequestHandler line 36
it uses the currentYearSubmission.CurrentSubmission.CompanyRegistrationNumber  which is always null

JohnPadgettCivica: @stevelee-civica OK lets remove that field from the database as its not needed and we shouldnt be updating it...

stevelee-civica: should I remove the CompanyRegistrationNumber column?

JohnPadgettCivica: Yeah remove that from DirectProducerSubmissionHistory ...  and from  the Domain class

PR #1589, opened 2024-10-31
76243 front

seanciv:


src/EA.Weee.Web/Views/OrganisationRegistration/ContactDetails.cshtml
```diff
@@ -70,6 +73,42 @@
...
         </div>
 
+        <div class="govuk-form-group @Html.Gds().FormGroupClass(m => m.AddressData.Postcode)">
+            @Html.Gds().LabelFor(m => m.AddressData.Postcode)
+            @Html.Gds().ValidationMessageFor(m => m.AddressData.Postcode)
+            @Html.Gds().TextBoxFor(m => m.AddressData.Postcode, new { @maxlength = CommonMaxFieldLengths.Postcode })
+        </div>
+
+        <button type="button"
+                name="Find address"
```
JohnPadgettCivica, line 83: @seanciv When JS is enabled a js-enabled attribute is added to the body tag. Wonder for those annoying people with JS disabled we hide the lookup fields by checking for existence of that attribute.  Would also check that WAVE is happy...  but looks good

seanciv: Just hid fields when js is disabled. WAVE seems fine @JohnPadgettCivica

PR #1606, opened 2024-11-06
Small producer feature flag

JohnPadgettCivica:


src/EA.Weee.Web/Controllers/OrganisationRegistrationController.cs
```diff
@@ -98,10 +98,20 @@ public async Task<ActionResult> Search(SearchViewModel viewModel)
...
+            DateTime currentDate;
+
+            using (var client = this.apiClient())
+            {
+                currentDate = await client.SendAsync(this.User.GetAccessToken(), new GetApiUtcDate());
+            }
+
+            var results = await organisationSearcher.Search(searchTerm, maximumSearchResults, false);
+
             SearchResultsViewModel viewModel = new SearchResultsViewModel
```
JohnPadgettCivica, line 110: @will-hope98  I would add a test for this as the code will be in place until it gets removed at some point in the future

will-hope98: I added tests on the OrgRegistrationControllerTests @JohnPadgettCivica

JohnPadgettCivica: Sorry :/

PR #1610, opened 2024-11-06
Company lookup UX enhancements

seanciv:


src/EA.Weee.Web/Scripts/lookup/org-details-company-lookup.js
```diff
@@ -0,0 +1,34 @@
...
+                if (response.HasError) {
+                    $errorBanner.show();
+                    $parent.addClass(errorClass)
+                } else {
+                    $errorBanner.hide();
+                    $parent.removeClass(errorClass)
+                }
+
+                if (response.Organisation) {
+                    $("#CompanyName").val(response.Organisation.Name)
```
JohnPadgettCivica, line 25: @seanciv  Looks good. Think this also needs to set the address fields as the server side version does.

PR #1613, opened 2024-11-07
78691 Back buttons in UJ1

stevelee-civica:
add back button

https://dev.azure.com/civica-digital/EA%20-%20WEEE/_workitems/edit/78691/

### DIFF
--- a/src/EA.Weee.Web/Views/OrganisationRegistration/JoinOrganisation.cshtml
+++ b/src/EA.Weee.Web/Views/OrganisationRegistration/JoinOrganisation.cshtml
@@ -1,7 +1,6 @@
 @using MvcNavigationHelpers
 @using EA.Weee.Web.RazorHelpers
 @using EA.Weee.Web.Controllers
-@using EA.Weee.Web.ViewModels.OrganisationRegistration
 @model EA.Weee.Web.ViewModels.OrganisationRegistration.JoinOrganisationViewModel
 @{
     ViewBag.Title = "Access Organisation";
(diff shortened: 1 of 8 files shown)


src/EA.Weee.Web/Views/OrganisationRegistration/JoinOrganisation.cshtml
```diff
@@ -1,10 +1,14 @@
+@using MvcNavigationHelpers
 @using EA.Weee.Web.RazorHelpers
+@using EA.Weee.Web.Controllers
 @model EA.Weee.Web.ViewModels.OrganisationRegistration.JoinOrganisationViewModel
 @{
     ViewBag.Title = "Access Organisation";
 }
+@section full_main_content 
+{
```
JohnPadgettCivica, line 9: @stevelee-civica  I dont think this will work. We need a way of determining where we have come from.
You can get here...
1. Login -> Access another org -> find existing org -> select -> continue , click back. We shouldn't go to organisationFound as its the search results screen that would have been the previous page in this scenario.  - This fails
2. Login -> Access another org -> Create small producer -> .... -> find company -> request to join -> back (This works ok)

JohnPadgettCivica:
@stevelee-civica 
Dont think we cover off 
UJ1: Producer details
Back button needs to take users back to 'Your organisation has been found' if they have come from that page

This is when you go through the process and have selected you are an auth rep. You enter org details  but there are similar, you click continue rather than selecting an existing organisation, on the producer details screen the back button should then take you to the organisation found screen.

src/EA.Weee.Web/Views/OrganisationRegistration/JoinOrganisation.cshtml
JohnPadgettCivica: @stevelee-civica  Actually looking at the AC im not sure this change is required at all. 
UJ1: Your organisation has been found
Add back button to take users back to Organisation details

Commented on the other. So you might be able to remove this change completely.

stevelee-civica: removed

stevelee-civica:
updated to use returnUrl 
the organisation-found page will set the returnUrl so RepresentingCompanyRedirect can use this for redirect

PR #1618, opened 2024-11-11
Resolve issue with starting a submission

JohnPadgettCivica:
https://dev.azure.com/civica-digital/EA%20-%20WEEE/_workitems/edit/78832
https://dev.azure.com/civica-digital/EA%20-%20WEEE/_workitems/edit/78833

These changes are mainly around moving the submission creation for the year into its own attribute so the yearly submission record is not created when view previous submissions or organisation details is accessed, its only created when doing the new/ continue submission option. As you can get into the organisation details screen without having a "current submission" fixed some null values.

Also makes a change so that when -1 as a year is passed to the submissions service the selected year is defaulted to the most recent but allows us to get into the 

image

Part of code when redirecting from choose activity if required.


src/EA.Weee.RequestHandlers/Organisations/OrganisationByIdHandler.cs
```diff
@@ -133,8 +122,7 @@ private async Task SetDirectRegistrants(OrganisationData organisationData, Guid
                     YearSubmissionStarted = directRegistrant.HasCurrentYearSubmission,
                     RepresentedCompanyName = directRegistrant.AuthorisedRepresentativeId.HasValue
                         ? directRegistrant.OverseasProducerName
-                        : null,
-                    MostRecentSubmittedYear = directRegistrant.MostRecentSubmittedYear
```
JohnPadgettCivica, line 137: No longer required as we just pass -1 from the Choose Activity to allow the registrations to select the most recent available year

PR #1625, opened 2024-11-13
78465-Show incomplete registrations to external users

will-hope98:
Added new continue registration button, now show all submissions to external

### DIFF
--- a/src/EA.Weee.Web/Areas/Producer/Controllers/ProducerController.cs
+++ b/src/EA.Weee.Web/Areas/Producer/Controllers/ProducerController.cs
@@ -1,10 +1,12 @@
 ﻿namespace EA.Weee.Web.Areas.Producer.Controllers
 {
+    using Api.Client;
     using EA.Prsd.Core;
     using EA.Prsd.Core.Mapper;
     using EA.Weee.Core;
     using EA.Weee.Core.Constants;
     using EA.Weee.Core.DirectRegistrant;
+    using EA.Weee.Requests.Shared;
     using EA.Weee.Web.Areas.Producer.Filters;
     using EA.Weee.Web.Areas.Producer.Mappings.ToViewModel;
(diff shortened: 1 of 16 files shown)


src/EA.Weee.Web/Areas/Producer/ViewModels/OrganisationDetailsTabsViewModel.cs
```diff
@@ -27,6 +28,8 @@ public class OrganisationDetailsTabsViewModel
         public bool IsRegistered => this.Status == SubmissionStatus.Submitted && this.HasPaid;
 
         public bool ShowReturnRegistrationToUser => this.IsInternalAdmin && (IsRegistered || this.Status == SubmissionStatus.Submitted);
+        public bool ShowContinueRegistrationToUser => !this.IsInternalAdmin && this.Status == SubmissionStatus.InComplete && this.ApiDateYear == this.Year;
```
JohnPadgettCivica, line 31: @will-hope98 probaby worth adding some unit tests for this property... think there is a similar test for ShowPaymentLink

src/EA.Weee.Web/Areas/Producer/Controllers/ProducerController.cs
```diff
@@ -162,6 +167,11 @@ public async Task<ActionResult> Submissions(int? year = null)
             submissionService.WithSubmissionData(this.SmallProducerSubmissionData);
 
             var model = await submissionService.Submissions(year);
+            using (var client = this.apiClient())
+            {
```
JohnPadgettCivica, line 171: @will-hope98 Do we need to do it for all of the tab actions in the ProducerController as I suspect if you switch tabs the button will vanish.  Maybe add a parameter to WithSubmissionData with the API year and set it in there, which would show the external places that it needs to be updated. Perhaps this could be a nullable int then it could be defaulted to null so any internal screens dont need a value passed.

PR #1651, opened 2024-12-05
Style cop update

JohnPadgettCivica:


CivicaJonathanStratford:
I don't know if it would work for this project (or what you're trying to achieve here), but for others we've had success with a `.editorconfig` file in the repo, which is set up to format files according to various code style rules -- and we've just fixed the code style issues as and when we next modify that file -- if we want to edit a file, we would open it, save it, so the style changes happen, commit that and then work on the actual changes we wanted -- so it's easy to just look at the changes. It does mean that files we don't edit never get fixed, but perhaps that's lower risk than touching lots of files (and presumably having to test all of the changes).

PR #1656, opened 2024-12-18
Changes for registered date in producer details

JohnPadgettCivica:
This change is to use the date of payment of a registration as the registered date rather than the date the submission was submitted.


sreedhar-civica (approved):
Looks good to me

PR #1662, opened 2025-01-09
Added Category 15 Vaping devices

sreedhar-civica:


CivicaJonathanStratford (changes requested):
The changes look fine as far as they go. However, I'm not sure if they're complete. There's at least one test failure which seems to be due to this change (I'm aware there are some that you said are unrelated, and fixed elsewhere), and I want to ensure that other changes that will be impacted by this are either covered in other tasks, or dealt with here. Unfortunately, as this is the first time that a WeeeCategory has been added from the history of the repository as far as I can tell, we'll have to do the work to ensure we find all the places.

- There is at least one test which check all of the other WeeeCategories, but hasn't been updated to add this one (GetUkWeeeAtAatfsCsvHandlerTests). I'm not sure if that one will be covered later, and I haven't check for more
- There is at least one stored procedure ([Evidence].[getSchemeObligationAndEvidenceTotals]) which expects only 1-14 for WeeeCategory; should that be updated here, or is is covered elsewhere?
- There's at least one class which expects WeeeCategory to be <= 14 -- AatfSummaryTotalsMap -- should that be updated here, or is that covered somewhere else?
- There are test failures in your build -- which I believe you explained were to do with dates in the past -- however, at least one of them appears to actually be due to the WeeeCategory -- EvidenceTonnageDataTests. So the failures need reviewing and fixing where it makes sense.

src/EA.Weee.Database/scripts/Everytime/0011-InsertWeeeCategory.sql
CivicaJonathanStratford, line 1: Although it's not your fault, the tblLocalArea variable here is confusing, can we rename it please?

CivicaJonathanStratford (changes requested):

src/EA.Weee.Database/scripts/Everytime/0011-InsertWeeeCategory.sql
sreedhar-civica: Better to leave the tblLocalArea variable as it is.

sreedhar-civica:
> The changes look fine as far as they go. However, I'm not sure if they're complete. There's at least one test failure which seems to be due to this change (I'm aware there are some that you said are unrelated, and fixed elsewhere), and I want to ensure that other changes that will be impacted by this are either covered in other tasks, or dealt with here. Unfortunately, as this is the first time that a WeeeCategory has been added from the history of the repository as far as I can tell, we'll have to do the work to ensure we find all the places.

> * There is at least one test which check all of the other WeeeCategories, but hasn't been updated to add this one (GetUkWeeeAtAatfsCsvHandlerTests). I'm not sure if that one will be covered later, and I haven't check for more.
**Fixed**

> * There is at least one stored procedure ([Evidence].[getSchemeObligationAndEvidenceTotals]) which expects only 1-14 for WeeeCategory; should that be updated here, or is is covered elsewhere?
**This will address during report changes**

> * There's at least one class which expects WeeeCategory to be <= 14 -- AatfSummaryTotalsMap -- should that be updated here, or is that covered somewhere else?
**This will be addressed during External user - Manage Evidence note screes**

> * There are test failures in your build -- which I believe you explained were to do with dates in the past -- however, at least one of them appears to actually be due to the WeeeCategory -- EvidenceTonnageDataTests. So the failures need reviewing and fixing where it makes sense.
**Fixed all failed tests**

src/EA.Weee.Database/scripts/Everytime/0011-InsertWeeeCategory.sql
CivicaJonathanStratford: OK -- though it's quite common to make small improvements like this when doing development work.

CivicaJonathanStratford (approved):
Having discussed this with you, I don't think I managed to explain my ideas very well, sorry. I'd meant for us to somehow make sure all examples were dealt with, rather than just the individual examples in my review points -- but you pointed out that things will get fixed as part of the different stories, and I guess if any fall through the cracks and don't get dealt with, they weren't necessary.

CivicaJonathanStratford (approved):
I think you could have been able to leave the Enum the same, which would have reduced the changes -- but since this hasn't gone in yet, it probably makes sense to have it with the right name to start with...so that's fine.

PR #1675, opened 2025-02-06
Add Vapes to pcs obligation upload template

CivicaJonathanStratford:


src/EA.Weee.Core/Shared/CsvReading/IWeeeCsvReader.cs
```diff
@@ -8,7 +8,7 @@ public interface IWeeeCsvReader : IDisposable
         void RegisterClassMap<T>() where T : ClassMap;
         bool Read();
         bool ReadHeader();
-        void ValidateHeader<T>(int expectedColumnCount, string[] expectedOrder);
```
sreedhar-civica, line 11: Could you please let me know the reason to delete the 'expectedColumnCount' parameter?

CivicaJonathanStratford: I wondered about it -- and if you want it put back, that's fine -- but it currently seems to be serving no purpose, because the way it is implemented the code is requiring you to provide the full set of columns so `expectedOrder` must have exactly the right number of elements (they're compared with `if (!csvReader.HeaderRecord.SequenceEqual(expectedOrder))`), as well as telling it how many columns there are.

I couldn't see from the git history why it was implemented this way -- it's possible a future implementation might want to let you say something like "there must be 10 columns, but I only want to specify the order / names of the first 5"), which my change prevents. But in that case, there's still a reasonable simplification to do which I should have done first -- in the callers, there's a constant defined for the array itself and for the length of the array, which seems to just be adding maintenance work because that length can just be passed in. I'd be surprised if the efficiency of having it in a constant is more important, but I might be wrong.

sreedhar-civica: Thank for the information. Just need to monitor if there are any selenium tests is going to fail related to this change. Approved.

sreedhar-civica (approved):

PR #1686, opened 2025-03-18
OMP - Internal User - All appropriate authority Issued charge changes

sreedhar-civica:


src/EA.Weee.Domain.Tests.Unit/Producer/ProducerSubmissionTests.cs
```diff
@@ -433,6 +436,58 @@ public void Producer_SetProducerMemberUpload_HasAnnualCharge_Returns_No()
...
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(3)]
+        public void ProducerMemberUpload_SellingTechniqueType_Returns_SellingTechniqueName_ChargeValue(int sellingTechniqueType)
+        {
+            // Arrange
+            var sellingType = new CustomSellingTechniqueType(sellingTechniqueType);
+            var sellingTypeName = EnumHelper.GetDisplayName(Enumeration.FromValue<SellingTechniqueType>(sellingTechniqueType));
+            decimal chargeThisUpdate = (sellingTechniqueType == 3) ? Convert.ToDecimal(9.999) : 0;
```
CivicaJonathanStratford, line 449: could just put an "m" on the end of 9.999 and 0 rather than call Convert.ToDecimal().

CivicaJonathanStratford, line 449: I think it might be better to have the have the decision whether to charge be part of the InlineData, rather than logic in the test

src/EA.Weee.Domain.Tests.Unit/Producer/ProducerSubmissionTests.cs
```diff
@@ -2,10 +2,13 @@
 {
     using Domain.Producer.Classfication;
     using Domain.Producer.Classification;
+    using EA.Prsd.Core.Domain;
+    using EA.Prsd.Core.Helpers;
     using EA.Weee.Domain.Producer;
     using EA.Weee.Domain.Scheme;
     using FakeItEasy;
     using Lookup;
+    using Microsoft.SqlServer.Server;
```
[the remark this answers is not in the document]
sreedhar-civica: Removed. :)

src/EA.Weee.Domain.Tests.Unit/Producer/ProducerSubmissionTests.cs
sreedhar-civica: passing as inline data now.

sreedhar-civica: passing as inline data now.

CivicaJonathanStratford: that's nice that the cast happens implicitly. However, that's not quite the change I had in mind

CivicaJonathanStratford: I wanted the *decision* to be passed in -- I don't think varying the value makes a useful difference, and it's not used when it's 1 or 2 anyway, is it?

sreedhar-civica: just to make sure for different selling technique types returns different charge values.

src/EA.Weee.Domain/Producer/ProducerSubmission.cs
```diff
@@ -217,6 +218,22 @@ public string DatabaseObligationType
...
+            {
+                return EnumHelper.GetDisplayName(Enumeration.FromValue<SellingTechniqueType>(SellingTechniqueType));
+            }
+        }
+
+        public decimal? OMPChargeValue
+        {
+            get
+            {
+                return (SellingTechniqueType == 3) ? ChargeThisUpdate : null; // 3 means Online Marketplaces and Fulfilment Houses
```
CivicaJonathanStratford, line 233: I'm not convinced this logic should be in the Domain

sreedhar-civica: As per my knowledge, if the Charge value stored in database in different column then will be implement in Domain. It's a dynamic value to get the data based on the 'Charge this value' column.

CivicaJonathanStratford: right, that's where I'm not sure -- the domain would just be doing what it's told in that case, as we'd be passing the relevant value in and getting it out, so it feels more like the caller should just decide to pass the same value in to the domain layer. Not sure though.

src/EA.Weee.Domain.Tests.Unit/Producer/ProducerSubmissionTests.cs
sreedhar-civica: I have checked yes it is using the passed values. please check.

src/EA.Weee.DataAccess/Mappings/ProducerSubmissionMapping.cs
```diff
@@ -20,6 +20,10 @@ public ProducerSubmissionMapping()
 
             Ignore(ps => ps.ObligationType);
             Property(ps => ps.DatabaseObligationType).HasColumnName("ObligationType");
+
+            Ignore(ps => ps.SellingTechniqueTypeName);
```
CivicaJonathanStratford, line 24: These properties are not derived so shouldn't be ignored, I don't think

src/EA.Weee.Domain/Producer/ProducerSubmission.cs
```diff
@@ -217,6 +217,12 @@ public string DatabaseObligationType
             set { ObligationType = (ObligationType)Enum.Parse(typeof(ObligationType), value); }
         }
 
+        public string SellingTechniqueTypeName { get; set; }
```
CivicaJonathanStratford, line 220: all these setters should be private

CivicaJonathanStratford: (only the setters)

src/EA.Weee.RequestHandlers/Charges/FetchIssuedChargesCsv/FetchIssuedChargesCsvHandler.cs
```diff
@@ -35,15 +39,27 @@ public async Task<FileInfo> HandleAsync(FetchIssuedChargesCsv message)
 
             IEnumerable<ProducerSubmission> results = await dataAccess.FetchInvoicedProducerSubmissionsAsync(authority, message.ComplianceYear, message.SchemeId);
 
+            if (results != null && results.Count() > 0)
```
CivicaJonathanStratford, line 42: this logic needs tests

src/EA.Weee.Domain/Producer/ProducerSubmission.cs
CivicaJonathanStratford, line 220: these properties need adding to the Equals method

CivicaJonathanStratford: so they'd need setting in the constructor

CivicaJonathanStratford:
I've ended up implementing some of this, so while I'm happy, and Sreedhar can't approve his own PR, it needs a comment from him to confirm he's happy (or that he's not!)

CivicaJonathanStratford:
Sreedhar made some suggestions so I made some further change; think it's all sorted now.

PR #1734, opened 2026-01-14
session timeout script changes.

sreedhar-civica:


src/EA.Weee.Web/Scripts/weee-session-timeout.js
```diff
@@ -1,93 +1,98 @@
...
+        sessionLogoutTimer = setTimeout(logout,
+            sessionTimeoutInSeconds * 1000
+        );
     }
 
     function showSessionWarning() {
-
-        clearTimeout(countdownTimer); // Stop any previous countdown
-        setTimeValue(warningTimeInSeconds - 1);
+        sessionStartTime = Date.now(); // reset start for warning countdown
```
ianalvaradoCivica, line 41: This confused me logically its correct but the name is misleading. Shouldnt it be warningStartTime?

src/EA.Weee.Web/Scripts/weee-session-timeout.js
```diff
@@ -1,93 +1,98 @@
...
         }, 1000);
     }
 
-    function closeSessionWarning() {
+    function setTimeValue(remainingTimeInSeconds) {
+        $("#prsd-timeout-countdown").html(formatTime(remainingTimeInSeconds));
+    }
 
-        clearTimeout(countdownTimer); // Stop countdown when modal is closed
+    function closeSessionWarning() {
```
ianalvaradoCivica, line 67: does this not clear only the countdownTimer and not the sessionWarningTimer and sessionLogoutTimer?

So if closeSessionWarning() is called without restarting the session, the logout timer keeps running in the background and will still fire.

if there is an X or and esc key it will be bugged

ianalvaradoCivica (changes requested):
Sorry Sreed if my comments make sense please can you address if not maybe demo could make it more understandable :D

src/EA.Weee.Web/Scripts/weee-session-timeout.js
ianalvaradoCivica: i think it only has stay logged in and close option right?

ianalvaradoCivica (approved):
Thanks for the demo Sreed. All good to go

PR #1744, opened 2026-02-17
Amendment Two - Remove the link ‘Download as PDF’ for ExternalUI.

vanessamartins2146:
- During the Direct Registrant user journey on the external UI there is a ‘Check your answers before submitting your application’ screen. At the bottom of this screen there is a link to ‘Download as PDF’.
-This ‘Download as PDF’ link is no longer needed and is removed.


vanessamartins2146:
Added further changes to this Amendment 2 section.

PR #1749, opened 2026-02-20
Add new Producer type of "Sole trader or individual" to applicable PC…

vanessamartins2146:
When a PCS member is registered as a partnership with >1 partners then their producer type is designated as a “Partnership”. This is correct.

However, when a PCS member is registered as a partnership with exactly 1 partner then their producer type is incorrectly designated as a “Partnership”. In this scenario they should be designated as a “Sole trader or individual”.

Any PCS member registrations going forward should apply the above logic.


vanessamartins2146 (commented):
Updated the comment and description section.

ianalvaradoCivica (approved):
the sql script are not appearing on the changes you made.  looks good

PR #1753, opened 2026-03-12
Inject ISystemDataDataAccess for system date retrieval

ianalvaradoCivica:
Updated EnvironmentAgencyProducerChargeBandCalculator to accept ISystemDataDataAccess via constructor and use it to obtain the current system date/time asynchronously. This replaces direct usage of DateTime.UtcNow, improving testability and allowing for configurable or mockable system time sources.


sreedhar-civica (approved):
Looks good to me.

PR #1754, opened 2026-03-13
Remove ISystemDataDataAccess dependency and update tests

ianalvaradoCivica:
EnvironmentAgencyProducerChargeBandCalculator no longer depends on ISystemDataDataAccess; system date is now obtained via DateTime.UtcNow.


sreedhar-civica (approved):
LGM

PR #1764, opened 2026-04-10
Inserted missing charge for year 2027. Added additional unit test.

ianalvaradoCivica:
Fixes from the bug that Matthew has raised during the demo last week.

When a submission compliance year is 2027 it doesnt return the correct figure for the following year.


sreedhar-civica (approved):
LGM:)

PR #1765, opened 2026-04-14
Update annual charge logic to support compliance year values

ianalvaradoCivica:
Updated Environment Agency annual charge seed data to 13,948.13. Refactored BySchemeTransactionFileGenerator to use IAnnualChargeDataAccess for retrieving annual charge amounts by compliance year, enabling support for year-specific values instead of a static property. Adjusted invoice line item descriptions accordingly.


sreedhar-civica (approved):
LGM :)

PR #1766, opened 2026-04-16
Refactor POST Index to redirect with filter parameters

ianalvaradoCivica:
Update ManageEvidenceNotesController so the POST Index action redirects to the GET Index action, preserving all filter and paging parameters in the URL. Introduce BuildFilterRouteValues to construct route values from the view model, supporting the Post/Redirect/Get pattern and improving usability by making filter states shareable and bookmarkable.


sreedhar-civica (approved):
LGM :)

